PR #4, opened 2022-11-21
Install ruby-vips gem

littleforest:
Adds Ruby bindings for the [libvips image processing library](https://libvips.github.io/libvips).


littleforest:
Hmm, maybe we will fall back to mini_magick for now. https://github.com/libvips/ruby-vips/issues/219

littleforest:
Going to close this for now. We don't seem to be having any issues uploading files on Heroku.

PR #11, opened 2022-11-29
Add route constraints depending on site subdomain

littleforest:
This will allow the sign in routes and the file_imports routes only to be accessed when on an admin subdomain. The routes will return a 404 if accessed on a non-admin subdomain.

These changes also add different root paths depending on whether the user is signed in or not, or whether on a non-admin subdomain.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,8 +5,15 @@ Rails 7, and PostgreSQL 14.
 ## Development setup
 1. Run `bin/setup`
 2. Install [libvips][libvips] for use with ActiveStorage
+4. To access the admin pages, you must modify your `/etc/hosts` file:
+   ```
+   # Added for ApprenticeshipStandards.org
+   127.0.0.1 admin.example.localhost
+   # End ApprenticeshipStandards.org additions
+   ```
 3. Start rails app: `bin/dev`. The application will be available
-   at http://localhost:3000.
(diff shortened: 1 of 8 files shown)


littleforest:
@jenlu33 if you have a chance before our meeting on Thursday, could you check out this branch and run through the instructions in the README (on this branch), and make sure these changes work for you? I don't want to merge them without you testing it out because it will mess up your being able to access the sign in page to style it.

This change will force you to view the sign in page and the internal `file_imports#index` page at http://admin.example.localhost:3000. When viewing http://localhost:3000, then only the standards_imports pages will be available.

Let me know when you have had a chance to test it out and I will merge it.

jenlu33:
> @jenlu33 if you have a chance before our meeting on Thursday, could you check out this branch and run through the instructions in the README (on this branch), and make sure these changes work for you? I don't want to merge them without you testing it out because it will mess up your being able to access the sign in page to style it.
> 
> This change will force you to view the sign in page and the internal `file_imports#index` page at http://admin.example.localhost:3000. When viewing http://localhost:3000, then only the standards_imports pages will be available.
> 
> Let me know when you have had a chance to test it out and I will merge it.

@littleforest I just checked it out and everything is working fine for me

PR #16, opened 2022-12-13
Notify user when new files uploaded

FerPerales:
* Send email to admin when a file is uploaded
* Update config/environments/development.rb with mailcatcher config
* Remove Rails mail preview
* Update README

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -157,6 +157,8 @@ GEM
     nio4r (2.5.8)
     nokogiri (1.13.9-arm64-darwin)
       racc (~> 1.4)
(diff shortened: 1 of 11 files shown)


littleforest (approved):
@jenlu33 @FerPerales thanks for finishing this, and for adding in the text version of the email. I modified the spec slightly so it would check all mail versions:

```
      mail.body.parts.each do |part|
        expect(part.body.encoded).to match si.name
        expect(part.body.encoded).to match si.email
        expect(part.body.encoded).to match si.organization
        expect(part.body.encoded).to match "10 files uploaded"
      end
```

PR #20, opened 2023-01-02
Fix flash modals not closing

FerPerales:
* Add flowbite to benefit from tailwind components
* Use data-dismiss-target from flowbite to close alerts
* Add flash modal specs

dismiss

Reference: https://app.asana.com/0/1203289004376659/1203589397331722/f


spec/views/layouts/alerts_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+      render partial: "layouts/alerts", locals: {notice: "Sample notice"}
+
+      expect(rendered).to have_css "[aria-label=Close]"
+      expect(rendered).to have_css "[role=button]"
+      expect(rendered).to have_css '[data-dismiss-target="#notice-flash"]'
+      expect(rendered).to have_css "div#notice-flash"
+      expect(rendered).to have_text "Sample notice"
+    end
+  end
+end
```
littleforest, line 27: Should this get moved to a system spec so we can actually test the closing of the alert? Or do you feel like this is sufficient?

littleforest (commented):
@FerPerales thanks so much for working on this! Since we are using import maps, we shouldn't also need to be installing anything with yarn in node_modules; we should be adding the js in the vendor directory:

```
./bin/importmap pin flowbite --download
```

However, if we are ever going to want to customize Tailwind significantly, then [we can't use Tailwind and import maps together](https://github.com/rails/tailwindcss-rails/issues/66), and will have to switch to a CSS compiler. Should we just stick with the import maps set up for now though, until we decide we need fancier CSS customization? Let me know what you think.

FerPerales:
> @FerPerales thanks so much for working on this! Since we are using import maps, we shouldn't also need to be installing anything with yarn in node_modules; we should be adding the js in the vendor directory:
> 
> ```
> ./bin/importmap pin flowbite --download
> ```
> 
> However, if we are ever going to want to customize Tailwind significantly, then [we can't use Tailwind and import maps together](https://github.com/rails/tailwindcss-rails/issues/66), and will have to switch to a CSS compiler. Should we just stick with the import maps set up for now though, until we decide we need fancier CSS customization? Let me know what you think.

We can keep importmap and see how it goes for us. I'll move this part

spec/views/layouts/alerts_spec.rb
FerPerales: @littleforest I updated a spec to test that the alert is dismissable

spec/system/users/sessions/new_spec.rb
```diff
@@ -28,6 +28,9 @@
       expect(page).to_not have_link "Logout"
       expect(page).to have_text "Invalid Email or password"
       expect(page).to have_css('div[role="alert"]')
+
+      find("svg[aria-label=Close]").click
+      expect(page).to_not have_no_content "Invalid Email or password"
```
littleforest, line 33: Can we change this away from a double negative, though I'm confused why it's working. Don't we want it to be:

```
expect(page).to_not have_content "Invalid Email or password"
```
?

FerPerales: @littleforest good catch. It must be `expect(page).to have_no_content "Invalid Email or password"`

https://til.magmalabs.io/posts/ea2ce20798-behavior-of-capybaras-havenocontent-when-expected-with-tonot-and-with-to

littleforest (commented):
@FerPerales is this working for your locally? I still can't close the modals 🤔

thoughtbot-boost-support:
@littleforest I'll double check

FerPerales:
@littleforest I'll double check

FerPerales:
@littleforest it's not working. According to the [docs](https://flowbite.com/docs/getting-started/rails/#install-flowbite), the setup using npm is the way to go if you are using tailwindcss-rails (which is our case). If we want to avoid using node/npm at all cost, I can create a simple Stimulus controller to close alerts. 

I like flowbite since it already provides common UI patterns that we may use in the future, but I also see this as an opportunity to introduce Stimulus and start shaping the way we are going to use it in this project

What are your thoughts?

littleforest:
> If we want to avoid using node/npm at all cost

We don't have to avoid it at all costs, I think I am just now running up against my Rails 7 knowledge. So far I've gotten away with importmaps + Stimulus. We will need to set up a JavaScript bundler for production, right?

littleforest:
> I'll create a Stimulus controller and get rid of flowbite in the meantime. If we need to introduce more complex components, we can evaluate in the future how to do it . Does it sound good?

This sounds fine if you are okay with not using Flowbite! If we want to do anything fancy with Tailwind we will need to set up esbuild or something anyway, so if you want to just go ahead and do that now, we can go that route too. Whichever is fine.

spec/system/users/sessions/new_spec.rb
littleforest: Cool I did not know about `have_no_content`!

littleforest (approved):
Looks great, thanks for fixing this and getting us started on Stimulus!

PR #24, opened 2023-01-04
Create task to update or create occupations with onet code

jenlu33:


app/models/onet_code.rb
```diff
@@ -0,0 +1,3 @@
+class OnetCode < ApplicationRecord
+  validates :name, :code, presence: true
```
littleforest, line 2: Let's add a uniqueness validation to this too. How about also on the database level.

littleforest (commented):
Looks good! Just minor validation suggestion.

db/schema.rb
```diff
@@ -56,6 +56,19 @@
     t.text "description"
     t.datetime "created_at", null: false
     t.datetime "updated_at", null: false
+    t.string "onet_code"
+    t.string "rapids_code"
+    t.integer "time_based_hours"
+    t.string "hybrid_hours"
+    t.integer "competency_based_hours"
+  end
```
littleforest, line 64: @jenlu33 can you remove this from the PR since it is actually part of the other PR with the rapids data import?

jenlu33: @littleforest Is there a way for me to get rid of the migration in the branch? Or do I just manually remove those lines? Since it always reappears when I run `rake db:migrate`

PR #26, opened 2023-01-06
Add task to scrape RAPIDS codes and create corresponding occupation columns

jenlu33:
https://app.asana.com/0/1203289004376659/1203641286036689/f


db/migrate/20230106195916_add_rapids_columns_to_occupation.rb
```diff
@@ -0,0 +1,8 @@
+class AddRapidsColumnsToOccupation < ActiveRecord::Migration[7.0]
+  def change
+    add_column :occupations, :rapids_code, :string
+    add_column :occupations, :time_based_hours, :integer
+    add_column :occupations, :hybrid_hours, :string
```
littleforest, line 5: @Pcushing @jenlu33 do we need to be storing the hybrid hours as either a range, or two separate fields with a min/max?

jenlu33: @littleforest I just changed the hybrid hours to an integer range and I also updated the task to set the range.

littleforest:
@jenlu33 can you fix merge conflicts and I can review?

lib/tasks/occupation.rake
```diff
@@ -1,6 +1,39 @@
...
+        hybrid_hours = nil
+
+        if row["HYBRID"]
+          hybrid_start_hours = row["HYBRID"].match(/(\d{1,}) -/)
+          hybrid_end_hours = row["HYBRID"].match(/- (\d{1,})/)
+        end
+
+        if hybrid_start_hours && hybrid_end_hours
+          hybrid_hours = (hybrid_start_hours[1]..hybrid_end_hours[1])
+        end
```
littleforest, line 22: @jenlu33 thanks for making the range change. Do you know if the Hybrid hours will always be a range? Otherwise we need to account for the case if it is just a single number, and set that for the start and end range.

jenlu33: @littleforest Sorry it took me a while to look back at this again. I tried googling, but I couldn't find any real answers. I added a range that just has the single number for both the starting and end numbers... but it seems like not the best solution?

littleforest: Yes, that's exactly right! I think this can be simplified a bit by putting all of this under the initial check for `if row["HYBRID"]` (and maybe give the `row["HYBRID"]` a variable name):

```
        if row["HYBRID"]
          hours = row["HYBRID"]
          hybrid_start_hours = hours.match(/(\d{1,})\s*-/)
          hybrid_end_hours = hours.match(/-\s*(\d{1,})/)

          if hybrid_start_hours && hybrid_end_hours
            hybrid_hours = (hybrid_start_hours[1]..hybrid_end_hours[1])
          else
            hybrid_hours = (hours..hours)
          end
        end
```

Also I added a `\s*` in the regex to have a little more flexibility in how the numbers come through (so it will match on "2000-4000" and "2000 - 4000".

littleforest (approved):
Looks great!

PR #28, opened 2023-01-10
Add list of registering apprenticeship agencies

littleforest:
https://app.asana.com/0/1203289004376659/1203641286036699/f

This adds a State model and a RegistrationAgency that is linked to a state. I added the [After Party gem](https://github.com/theSteveMitchell/after_party) to help with one-time data imports.

The file used for state names and abbreviations was taken from: https://www2.census.gov/geo/docs/reference/state.txt

I created the file used for the registration_agencies table manually from the references in the ticket.


app/models/registration_agency.rb
```diff
@@ -0,0 +1,12 @@
+class RegistrationAgency < ApplicationRecord
+  validates :agency_type, presence: true
+  validates :state, uniqueness: {scope: :agency_type}
+
+  belongs_to :state
```
FerPerales, line 5: Do we want to use `shoulda_matcher` to test that any validation and association is in place? We haven't discuss about the value that provides, but we use it for other clients

lib/tasks/deployment/20230109153235_populate_states.rake
```diff
@@ -0,0 +1,21 @@
+require "csv"
+
+namespace :after_party do
+  desc "Deployment task: populate_states"
+  task populate_states: :environment do
+    puts "Running deploy task 'populate_states'"
+
+    filename = File.join(Rails.root, "lib", "tasks", "files", "state.txt")
```
FerPerales, line 8: nit: we could rename the file to `state.csv`, since it seems to be a csv and not a txt

FerPerales (approved):
Looks good. Left a couple of comments

app/models/registration_agency.rb
littleforest: I always feel like `shoulda_matcher` is a little overkill and I don't use it, but I can be persuaded if others find value in it.

lib/tasks/deployment/20230109153235_populate_states.rake
littleforest: Well, it is not comma-separated, and doesn't actually open nicely in Numbers, so I figured it was better to leave it as `.txt`. It seems that the CSV library is smart enough to handle it, but I don't think it's technically a CSV file?

PR #31, opened 2023-01-11
Destroy file_import when standards_import or attachment is destroyed

littleforest:
This makes sure to destroy the file_import record that is associated to the ActiveStorage::Attachment record when either the attachment is removed directly, or when the StandardsImport record itself is removed.

### DIFF
--- a/spec/models/standards_import_spec.rb
+++ b/spec/models/standards_import_spec.rb
@@ -7,6 +7,23 @@ RSpec.describe StandardsImport, type: :model do
     expect(si).to be_valid
   end
 
+  it "deletes file import record when deleted" do
+    standards_import = create(:standards_import, :with_files)
+    attachment = standards_import.files.first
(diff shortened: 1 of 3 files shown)


spec/models/standards_import_spec.rb
```diff
@@ -7,6 +7,23 @@
     expect(si).to be_valid
   end
 
+  it "deletes file import record when deleted" do
+    standards_import = create(:standards_import, :with_files)
+    attachment = standards_import.files.first
+    file_import = attachment.file_import
+
+    expect(attachment).to be
```
FerPerales, line 15: Do we need something else on these expectations?

littleforest: No it is basically saying that it expects it to exist. I just did a quick test where I set `foo = nil` and got:

```
Failure/Error: expect(foo).to be
       expected nil to evaluate to true
```

PR #34, opened 2023-01-13
Create Data Import

jenlu33:
Asana ticket: [https://app.asana.com/0/1203289004376659/1203289004376671/f](https://app.asana.com/0/1203289004376659/1203289004376671/f)

### DIFF
--- a/app/models/data_import.rb
+++ b/app/models/data_import.rb
@@ -9,7 +9,7 @@ class DataImport < ApplicationRecord
 
   def file_presence
     unless file.attached?
-      errors.add(:file, "Must upload file")
+      errors.add(:file, "must be attached")
     end
   end
 end
--- a/app/models/data_import.rb
+++ b/app/models/data_import.rb
@@ -1,0 +1,6 @@
+class DataImport < ApplicationRecord
+  has_one_attached :file
+  
+  belongs_to :user
+
+end
--- a/app/models/data_import.rb
+++ b/app/models/data_import.rb
@@ -12,5 +12,4 @@ class DataImport < ApplicationRecord
       errors.add(:file, "Must upload file")
     end
   end
-
 end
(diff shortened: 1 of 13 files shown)


app/models/data_import.rb
```diff
@@ -0,0 +1,15 @@
...
+
+  validate :file_presence
+
+  private
+
+  def file_presence
+    unless file.attached?
+      errors.add(:file, "Must upload file")
+    end
+  end
```
littleforest, line 14: I tested this in the UI, and it doesn't seem that our file validation is working.

app/views/data_imports/show.html.erb
```diff
@@ -0,0 +1,20 @@
...
+      <div class="show-label">File</div>
+        <div class="w-full mb-1">
+          <%= link_to @data_import.file.filename.to_s, url_for(@data_import.file), class: "text-blue-600 hover:text-blue-800 visited:text-purple-600" %>
+        </div>
+    </div>
+  </div>
+  <div class=" max-w-prose mx-auto flex text-center space-x-2">
+    <%= link_to "Upload more standards", new_data_import_path, class: "btn-primary" %>
+    <%= link_to "Edit occupation standard", edit_data_import_path(@data_import), class: "btn-primary" %>
+    <%= button_to "Delete occupation standard", data_import_path(@data_import), method: :delete, data: {turbo_method: :delete}, class: "btn-primary" %>
```
littleforest, line 18: Ah nice, you got this to work!

littleforest (commented):
@jenlu33 this is looking great! Could we round it out with adding the index method and adding a link in the navbar?

app/models/data_import.rb
jenlu33: @littleforest I added the error message into the views and it looks like the data import is not being created unless a file is attached

PR #35, opened 2023-01-17
Create Work Process

jenlu33:
Asana ticket: [https://app.asana.com/0/1203289004376659/1203683433495313/f](https://app.asana.com/0/1203289004376659/1203683433495313/f)

This just creates the work process model for now. I was going to create a controller and views, but realized that depends on the data import and the views might not be necessary.


db/migrate/20230116181439_create_work_processes.rb
```diff
@@ -0,0 +1,15 @@
+class CreateWorkProcesses < ActiveRecord::Migration[7.0]
+  def change
+    create_table :work_processes, id: :uuid do |t|
+      t.string :title
+      t.string :description
+      t.references :occupation_standard, null: false, foreign_key: true, type: :uuid
+      t.integer :default_hours
+      t.integer :minimum_hours
+      t.integer :maximum_hours
```
littleforest, line 9: @jenlu33 could you make this a range like you did for the `hybrid_hours` field on the occupations table? That way we can more easily query work processes that fall within a given search range.

db/schema.rb
```diff
@@ -137,11 +137,25 @@
     t.index ["reset_password_token"], name: "index_users_on_reset_password_token", unique: true
   end
 
+  create_table "work_processes", id: :uuid, default: -> { "gen_random_uuid()" }, force: :cascade do |t|
+    t.string "title"
+    t.string "description"
+    t.uuid "occupation_standard_id", null: false
+    t.integer "minimum_hours"
+    t.integer "maximum_hours"
```
littleforest, line 145: Sorry I missed these hours -- could these also be a range?

littleforest:
Also, please fix merge conflicts.

db/schema.rb
```diff
@@ -138,6 +139,19 @@
...
+  create_table "work_processes", id: :uuid, default: -> { "gen_random_uuid()" }, force: :cascade do |t|
+    t.string "title"
+    t.string "description"
+    t.uuid "occupation_standard_id", null: false
+    t.integer "sort_order"
+    t.datetime "created_at", null: false
+    t.datetime "updated_at", null: false
+    t.int4range "default_hours"
+    t.int4range "minimum_hours"
+    t.int4range "maximum_hours"
```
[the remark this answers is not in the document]
jenlu33: Oh that makes a lot more sense! I'll work on getting that changed around.

db/migrate/20230117191958_remove_min_and_max_hours_from_work_process.rb
```diff
@@ -0,0 +1,6 @@
+class RemoveMinAndMaxHoursFromWorkProcess < ActiveRecord::Migration[7.0]
+  def change
+    remove_column :work_processes, :minimum_hours
+    remove_column :work_processes, :maximum_hours
+  end
+end
```
littleforest, line 6: Let's just delete this migration

db/migrate/20230117192226_add_min_and_max_hours_to_work_process.rb
```diff
@@ -0,0 +1,6 @@
+class AddMinAndMaxHoursToWorkProcess < ActiveRecord::Migration[7.0]
+  def change
+    add_column :work_processes, :minimum_hours, "int4range"
+    add_column :work_processes, :maximum_hours, "int4range"
+  end
+end
```
littleforest, line 6: and also delete this migration

littleforest (commented):
@jenlu33 looks good! Can you just remove the 2 extraneous migrations?

db/migrate/20230117191958_remove_min_and_max_hours_from_work_process.rb
littleforest: @jenlu33 can you remove this file too? Otherwise the next migration will break.

db/migrate/20230117185953_add_default_hours_to_work_process.rb
```diff
@@ -1,5 +0,0 @@
-class AddDefaultHoursToWorkProcess < ActiveRecord::Migration[7.0]
-  def change
-    add_column :work_processes, :default_hours, "int4range"
-  end
-end
```
littleforest: Oops, this file should stay.

db/migrate/20230117191958_remove_min_and_max_hours_from_work_process.rb
littleforest: But looks like you removed `db/migrate/20230117185953_add_default_hours_to_work_process.rb` which should be reinstated.

littleforest (commented):
@jenlu33 sounds like the preference is to keep the `_min` and `_max` columns since we don't need to do any querying across these fields. It might be easiest if you reinstate the two files you removed in f8467fbef365cc88821c7cdce97c3e1698322017, and then do a `rails db:rollback STEP=7`. Then from there remove any unneeded migration files, and then migrate back up. Happy to pair if this is confusing. And check the count for `STEP=7` in case I counted incorrectly.

FerPerales (approved):

littleforest (approved):
🎉 sorry for all the back and forth!

PR #37, opened 2023-01-19
Create Competency model

FerPerales:
Asana ticket: https://app.asana.com/0/1203289004376659/1203683433495315

This create Competency model, which is associated to a WorkProcess

PR #38, opened 2023-01-19
Create Related Instruction Model

jenlu33:
Asana ticket: [https://app.asana.com/0/1203289004376659/1203683433495319/f](https://app.asana.com/0/1203289004376659/1203683433495319/f)

### DIFF
--- a/app/models/related_instruction.rb
+++ b/app/models/related_instruction.rb
@@ -1,3 +1,5 @@
 class RelatedInstruction < ApplicationRecord
   belongs_to :occupation_standard
+
+  validates :sort_order, uniqueness: {scope: :occupation_standard}
 end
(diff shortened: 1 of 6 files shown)


db/migrate/20230118205341_create_related_instructions.rb
```diff
@@ -0,0 +1,14 @@
+class CreateRelatedInstructions < ActiveRecord::Migration[7.0]
+  def change
+    create_table :related_instructions, id: :uuid do |t|
+      t.string :title
+      t.int4range :hours
```
littleforest, line 5: @jenlu33 let's just make this a regular integer.

littleforest (commented):
Looks good! Just one field change in the db.

app/models/related_instruction.rb
```diff
@@ -0,0 +1,3 @@
+class RelatedInstruction < ApplicationRecord
+  belongs_to :occupation_standard
```
littleforest, line 2: @jenlu33 can you also make sort order unique within the occupation standard? See Fer's PR #37 in the `app/models/competency.rb` for similar set up.

littleforest (approved):
🎉

PR #39, opened 2023-01-20
Create Wage Step Model

jenlu33:
Asana ticket: [https://app.asana.com/0/1203289004376659/1203683433495323/f](https://app.asana.com/0/1203289004376659/1203683433495323/f)

### DIFF
--- a/app/models/wage_step.rb
+++ b/app/models/wage_step.rb
@@ -1,0 +1,3 @@
+class WageStep < ApplicationRecord
+  belongs_to :occupation_standard
+end
--- a/app/models/wage_step.rb
+++ b/app/models/wage_step.rb
@@ -1,3 +1,5 @@
 class WageStep < ApplicationRecord
   belongs_to :occupation_standard
+
+  validates :sort_order, uniqueness: {scope: :occupation_standard}
 end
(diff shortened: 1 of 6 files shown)


app/models/wage_step.rb
```diff
@@ -0,0 +1,3 @@
+class WageStep < ApplicationRecord
+  belongs_to :occupation_standard
```
littleforest, line 2: Similar to #37 and #38 can you make the sort order unique within the occupation standard?

db/schema.rb
```diff
@@ -99,6 +99,18 @@
...
+    t.string "title"
+    t.int4range "hours"
+    t.boolean "elective"
+    t.integer "sort_order"
+    t.uuid "occupation_standard_id", null: false
+    t.integer "default_course_id"
+    t.datetime "created_at", null: false
+    t.datetime "updated_at", null: false
+    t.index ["occupation_standard_id"], name: "index_related_instructions_on_occupation_standard_id"
+  end
```
littleforest, line 112: Could you remove the related_instructions piece out of the schema (including foreign key statement at the end).

littleforest (commented):
Looks good, just one request and some cleanup. Also, looks like some file linting needs to be fixed.

littleforest (approved):
🎉 looks great!

PR #40, opened 2023-01-23
Create Course Model

jenlu33:
Asana ticket: [https://app.asana.com/0/1203289004376659/1203683433495321/f](https://app.asana.com/0/1203289004376659/1203683433495321/f)\
I took out the `organization_id` for now since we don't have that model yet.

I also added in the reference to `course_id` on `RelatedInstruction` based on this previous ticket since we didn't have the course yet: [https://app.asana.com/0/1203289004376659/1203683433495319/f](https://app.asana.com/0/1203289004376659/1203683433495319/f)

### DIFF
--- a/db/migrate/20230123163754_create_courses.rb
+++ b/db/migrate/20230123163754_create_courses.rb
@@ -1,0 +1,14 @@
+class CreateCourses < ActiveRecord::Migration[7.0]
+  def change
+    create_table :courses, id: :uuid do |t|
+      t.string :title
+      t.string :description
+      t.string :code
+      t.decimal :units
+      t.integer :hours
+      t.integer :organization_id
+
+      t.timestamps
+    end
+  end
+end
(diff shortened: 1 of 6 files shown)


littleforest (approved):
Looks great!

db/migrate/20230123163754_create_courses.rb
```diff
@@ -0,0 +1,13 @@
+class CreateCourses < ActiveRecord::Migration[7.0]
+  def change
+    create_table :courses, id: :uuid do |t|
+      t.string :title
```
FerPerales, line 4: @jenlu33 do we need validation for any of these fields?

FerPerales (approved):

db/migrate/20230123163754_create_courses.rb
jenlu33: @FerPerales I think the `title` should be unique with the `organization_id`, but I can add that in after creating an organization model.

PR #41, opened 2023-01-23
Create competency option

FerPerales:
Asana ticket: https://app.asana.com/0/1203289004376659/1203683433495317/f

This PR creates CompetencyOption as a polymorphic table to support Competency and Occupation


app/models/competency.rb
```diff
@@ -2,4 +2,6 @@ class Competency < ApplicationRecord
   belongs_to :work_process
 
   validates :sort_order, uniqueness: {scope: :work_process}
+
+  has_many :competency_options, as: :resource
```
littleforest, line 6: Just for organization's sake, could we move the `has_many` statement up so the `belongs_to` and `has_many` are all grouped together and not broken up by the `validates statement?

littleforest (approved):
Looks great! I like the polymorphic set up. Just a minor line rearrangement comment.

PR #42, opened 2023-01-23
Fix db schema

littleforest:
Fixing the db schema so it matches what you get when starting from a freshly created database.


FerPerales (approved):

FerPerales:
Thanks for working on this! It was very annoying

PR #44, opened 2023-01-24
Create Organization model

jenlu33:
Asana ticket: [https://app.asana.com/0/1203289004376659/1203762334405088/f](https://app.asana.com/0/1203289004376659/1203762334405088/f)

### DIFF
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -1,2 +1,5 @@
 class Course < ApplicationRecord
+  belongs_to :organization
+
+  validates :title, :code, uniqueness: {scope: :organization}
 end
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -1,5 +1,5 @@
 class Course < ApplicationRecord
-  belongs_to :organization
+  belongs_to :organization, optional: true
 
   validates :title, :code, uniqueness: {scope: :organization}
 end
--- a/app/models/occupation_standard.rb
+++ b/app/models/occupation_standard.rb
@@ -1,7 +1,7 @@
 class OccupationStandard < ApplicationRecord
   belongs_to :occupation, optional: true
   belongs_to :registration_agency
-  belongs_to :organization
+  belongs_to :organization, optional: true
 
   has_many :related_instructions
   has_many :wage_steps
--- a/app/models/occupation_standard.rb
+++ b/app/models/occupation_standard.rb
@@ -1,6 +1,7 @@
 class OccupationStandard < ApplicationRecord
   belongs_to :occupation, optional: true
   belongs_to :registration_agency
+  belongs_to :organization
 
   has_many :related_instructions
   has_many :wage_steps
--- a/app/models/organization.rb
+++ b/app/models/organization.rb
@@ -1,5 +1,5 @@
 class Organization < ApplicationRecord
-  has_one_attached :file
+  has_one_attached :logo
 
   has_many :courses
   has_many :occupation_standards
--- a/app/models/organization.rb
+++ b/app/models/organization.rb
@@ -4,5 +4,5 @@ class Organization < ApplicationRecord
   has_many :courses
   has_many :occupation_standards
 
-  validates :title, presence: true
+  validates :title, presence: true, uniqueness: true
 end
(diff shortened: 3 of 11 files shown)


db/schema.rb
```diff
@@ -120,6 +124,13 @@
     t.index ["code"], name: "unique_code", unique: true
   end
 
+  create_table "organizations", id: :uuid, default: -> { "gen_random_uuid()" }, force: :cascade do |t|
+    t.string "title"
+    t.string "type"
```
littleforest, line 129: `type` is reserved for Single Table Inheritance. I often use `kind` instead, but whatever else sounds good to you (`organization_type` is a possibility too).

spec/factories/occupation_standards.rb
```diff
@@ -3,5 +3,6 @@
     occupation
     url { "http://example.com" }
     registration_agency
+    organization
```
littleforest, line 6: Since an organization is not always required for an occupation standard, let's leave this off the factory. In general factories should just contain the minimal fields required. It will be helpful to keep our test suite running fast by not having it create extra organization records every time it creates an occupation standard record.

spec/factories/courses.rb
```diff
@@ -4,5 +4,6 @@
     description { "Brake systems description" }
     code { "AA11" }
     units { "3.5" }
+    organization
```
littleforest, line 7: Same here -- is organization required? From the ticket it sounds like it can be optional, so let's leave it out to avoid creating unnecessary db records during the test suite.

app/models/occupation_standard.rb
```diff
@@ -1,6 +1,7 @@
 class OccupationStandard < ApplicationRecord
   belongs_to :occupation, optional: true
   belongs_to :registration_agency
+  belongs_to :organization
```
littleforest, line 4:
```suggestion
  belongs_to :organization, optional: true
```

app/models/course.rb
```diff
@@ -1,2 +1,5 @@
 class Course < ApplicationRecord
+  belongs_to :organization
```
littleforest, line 2:
```suggestion
  belongs_to :organization, optional: true
```

littleforest (commented):
Thanks for working on this! I added a dummy organization model in my PR but will remove it.

spec/factories/courses.rb
littleforest: I was just going off of the comment in the ticket: "**Most** of our courses will have a school of some sort attached to them" (bold mine). @Pcushing do you have input?

app/models/organization.rb
```diff
@@ -0,0 +1,8 @@
+class Organization < ApplicationRecord
+  has_one_attached :file
+
+  has_many :courses
+  has_many :occupation_standards
+
+  validates :title, presence: true
```
littleforest, line 7: Let's also add uniqueness validation here.
```suggestion
  validates :title, presence: true, uniqueness: true
```

spec/factories/organizations.rb
```diff
@@ -0,0 +1,5 @@
+FactoryBot.define do
+  factory :organization do
+    title { "Computer Academy" }
```
littleforest, line 3: Since we'll want a unique title, we can use a sequence in the factory definition:

```suggestion
    sequence(:title) { |n| "Computer Academy #{n}" }
```

app/models/organization.rb
```diff
@@ -0,0 +1,8 @@
+class Organization < ApplicationRecord
+  has_one_attached :file
```
littleforest, line 2: Can we name this `logo` so it's more clear what kind of file it is?

littleforest (commented):
@jenlu33 looks great, just one minor change requested for the attachment name.

app/models/organization.rb
jenlu33: @littleforest I renamed file to logo, but is there anything else I need to add? How does `Organization` know what `logo` is referring to?

littleforest: It doesn't really matter what the name is, it just makes for readability in the code. So you might have something like:  `if organization.logo.attached?`. It is just a little more clear what type of attachment we are expecting.

littleforest (approved):
🎉

PR #45, opened 2023-01-24
Configure Sidekiq

littleforest:
Asana ticket: https://app.asana.com/0/1203289004376659/1203812543964486/f

This sets up Sidekiq to be used in production on Heroku, as well as the Sidekiq dashboard.

Screen Shot 2023-01-24 at 2 45 55 PM

Screen Shot 2023-01-24 at 2 46 34 PM

I selected the Upstash Redis add-on as it gave the most for $0. Heroku Redis no longer has a free plan. We can modify this easily later if needed.

Followed the [Heroku setup guide](https://github.com/mperham/sidekiq/wiki/Heroku).

### DIFF
--- a/Procfile
+++ b/Procfile
@@ -1,2 +1,3 @@
 web: bundle exec puma -C config/puma.rb
+worker: bundle exec sidekiq -c 8
 release: bundle exec rails db:migrate && rake after_party:run
--- a/Gemfile
+++ b/Gemfile
@@ -44,6 +44,7 @@ gem "rollbar"
 gem "roo", "~> 2.9.0"
 gem "aws-sdk-s3", require: false
 gem "after_party"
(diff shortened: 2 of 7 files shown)


FerPerales (approved):

Procfile
```diff
@@ -1,2 +1,3 @@
 web: bundle exec puma -C config/puma.rb
+worker: bundle exec sidekiq -c 8
```
FerPerales, line 2: @littleforest do we also want this in `Procfile.dev` or just for prod?

littleforest: Good question! Does Rails just use inline for dev environment by default? If so, should we stick with that for the short-term so devs don't need to get Redis installed?

FerPerales: Sounds good!

PR #46, opened 2023-01-26
Add occupation standard list with related instructions

FerPerales:
Asana ticket: https://app.asana.com/0/1203289004376659/1203289004376677/f

This PR creates an index view for Occupation Standards:

image

image

Each record contains a link to Occupation Standard show view. In that view, you can also see the list of Related Instructions for the current Occupation Standard

image


littleforest (commented):
@FerPerales thanks for working on this! Do you think it's possible to use divs instead of a table for the index view so we can have better readability on small devices? Could it drop down to a vertical list on a small device?

Thanks for taking advantage of quicker spec views as well!

FerPerales:
@littleforest yes. I'll change it

littleforest:
On the other hand, we could leave as-is with a table and worry about it later in case a designer comes on. I'm fine either way.

FerPerales:
@littleforest based on discussion during sync, I'd go for keeping it since it's not the final design

littleforest (approved):
Works for me!

PR #48, opened 2023-01-31
Change hybrid hours to separate min/max fields

littleforest:
Asana ticket: https://app.asana.com/0/1203289004376659/1203831316199916/f

This removes the Postgres range field type for hybrid_hours and replaces it with min/max fields instead. It also updates the occupation scraper rake task to use the new fields.

### DIFF
--- a/lib/tasks/occupation.rake
+++ b/lib/tasks/occupation.rake
@@ -10,19 +10,19 @@ namespace :occupation do
         next if index.zero?
         occupation = Occupation.find_or_initialize_by(name: row["RAPIDS TITLE"])
         onet_code = OnetCode.find_by(code: row["ONET SOC CODE"].strip)
-        hybrid_start_hours = nil
-        hybrid_end_hours = nil
-        hybrid_hours = nil
+        hybrid_hours_min = nil
+        hybrid_hours_max = nil
 
         if row["HYBRID"]
           hours = row["HYBRID"]
-          hybrid_start_hours = hours.match(/(\d{1,})\s*-/)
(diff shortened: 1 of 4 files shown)


lib/tasks/occupation.rake
```diff
@@ -10,27 +10,28 @@ namespace :occupation do
...
-        hybrid_end_hours = nil
-        hybrid_hours = nil
+        hybrid_hours_min = nil
+        hybrid_hours_max = nil
 
         if row["HYBRID"]
           hours = row["HYBRID"]
-          hybrid_start_hours = hours.match(/(\d{1,})\s*-/)
-          hybrid_end_hours = hours.match(/-\s*(\d{1,})/)
+          hybrid_hours = hours.match(/(\d{1,})\s*-\s*(\d{1,})/)
```
FerPerales, line 18: **nit**: I'm a big fan of named captures, that allow us to assign a meaning name to every capture group in the regex

```ruby
hybrid_hours = /(?<hybrid_hours_min>\d{1,})\s*-\s*(?<hybrid_hours_max>\d{1,})/
if hybrid_hours
  hybrid_hours_min = hybrid_hours[:hybrid_hours_min]
  hybrid_hours_max = hybrid_hours[:hybrid_hours_max]
else
...
```

FerPerales (approved):

lib/tasks/occupation.rake
littleforest: Oh cool, I didn't even know about this! I will update.

PR #52, opened 2023-02-01
Related instructions import

jenlu33:
Asana ticket: [https://app.asana.com/0/1203289004376659/1203683433495299/f](https://app.asana.com/0/1203289004376659/1203683433495299/f)


app/services/import_occupation_standard_related_instruction.rb
```diff
@@ -0,0 +1,43 @@
...
+        organization = Organization.find_or_create_by!(
+          title: row["Related Training Organization"]
+        )
+
+        course = Course.where(
+          code: row["Course Code"],
+          organization: organization
+        ).first_or_create!(
+          description: row["Course Description"]
+        )
```
littleforest, line 28: @jenlu33 I thought that all of these course fields were supposed to actually be on the Related Instruction model? At our meeting before last, I was pretty sure the understanding was that this information wasn't about a specific course in our database but rather general information that would be stored on the Related Instruction record.

spec/factories/related_instructions.rb
```diff
@@ -5,5 +5,6 @@
     elective { false }
     sort_order { 1 }
     occupation_standard
+    association :default_course, factory: :course
```
littleforest, line 8: Let's leave this out of the factory since it's not required. The specs run faster when fewer database records are added, and we don't want to unnecessarily be adding a course record every time we create a RelatedInstruction object.

app/services/import_occupation_standard_related_instruction.rb
jenlu33: Some of the fields like code, description, and organization aren't on the Related Instruction model so I wasn't sure where else to store that information.

littleforest: I think we need to add the fields to the RelatedInstruction model.

jenlu33:
@littleforest I added the related instructions array back so the import file's specs could run correctly

littleforest (approved):
🎉 🚢

PR #56, opened 2023-02-07
Display occupation standard on data import page plus other improvements

littleforest:
Asana ticket: https://app.asana.com/0/1203289004376659/1203289004376673/f

I will seek clarification on this ticket, but we do not really need to add a UI action to associate a data import with an occupation standard since that will happen when the data import is parsed. In a separate PR I will link up that parsing with the uploading of a data import. For now, this PR makes the following changes:

* Replace the h2 tags with h1 tags for better accessibility.
* Adds a "cancel" button to some of the form pages. It doesn't have exactly the same height as the Submit button, but not worrying about this for now.
* Adds an "Upload data import" button to the data_imports#index page, as there wasn't any way to get to the new upload page.
* Add a link to the data_imports#show page from the index page.
* Changed some of the verbiage on the Data Import CRUD pages to reference "data import" instead of "occupation standard" since we have separate CRUD pages for Occupation Standard.
* Added a link to the Occupation Standard from the Data Import show page. Right now this is conditional on the presence of the occupation standard, but ultimately we should always have an occupation standard linked to a data import.
* Display the Occupation Standard title instead of only the Occupation title on the occupation_standards#show page. It is possible for an occupation standard not to be linked to an occupation.

Screen Shot 2023-02-07 at 11 56 46 AM

Screen Shot 2023-02-07 at 11 55 14 AM

Screen Shot 2023-02-07 at 11 55 21 AM

Screen Shot 2023-02-07 at 11 54 35 AM
Screen Shot 2023-02-07 at 11 53 58 AM

### DIFF
--- a/app/assets/stylesheets/application.tailwind.css
+++ b/app/assets/stylesheets/application.tailwind.css
@@ -6,6 +6,10 @@
   .btn-primary {
     @apply bg-blue-500 hover:bg-blue-700 text-white font-bold py-2 px-4 rounded
   }
+
+  .btn-secondary {
+    @apply bg-gray-500 hover:bg-gray-700 text-white font-bold py-2 px-4 rounded
+  }
 }
 
 @layer base {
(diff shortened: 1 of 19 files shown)


FerPerales (approved):
:shipit:

PR #57, opened 2023-02-08
Process data_import file after upload

littleforest:
Asana ticket: https://app.asana.com/0/1203289004376659/1203289004376673/f

This calls the background job that parses the data import file after a record is created. Calling this in the controller instead of as a callback on the model to avoid some circular issues.


littleforest:
@jenlu33 @FerPerales this is ready for review.

FerPerales (approved):
:shipit:

PR #58, opened 2023-02-08
scrape California's public posting of new programs or updated standards

littleforest:
Asana ticket: https://app.asana.com/0/1203289004376659/1203778503386690/f

This adds a background job to scrape the standards from: https://www.dir.ca.gov/das/ProgramStandards.htm

One new StandardsImport record is created using the name of the file as the name of the record to ensure that we do not re-download the same files every time the job is run.

This also adds the webmock gem to make sure we do not make external HTTP requests in tests.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -84,4 +84,5 @@ group :test do
   gem "capybara_accessible_selectors", git: "https://github.com/citizensadvice/capybara_accessible_selectors", tag: "v0.10.0"
   gem "selenium-webdriver"
   gem "webdrivers"
+  gem "webmock"
 end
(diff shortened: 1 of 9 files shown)


jenlu33:
@littleforest This looks good to me! Should I use this as a template to do the same for the other state websites that need to be scraped?

littleforest:
> Should I use this as a template to do the same for the other state websites that need to be scraped?

Sure, unless you have some other ideas that you were thinking of! There are probably lots of ways to handle this.

FerPerales (approved):
Looks good! We could probably extract some common functionality to a base class once we have more scrappers, but this is a very great start!

PR #62, opened 2023-02-14
Add ImportOccupationStandardWorkProcesses service

FerPerales:
Asana ticket: https://app.asana.com/0/1203289004376659/1203683433495297/f

This PR creates WorkProcess and Competency objects from a data import


app/services/import_occupation_standard_work_processes.rb
```diff
@@ -0,0 +1,61 @@
...
+      competency.assign_attributes(
+        title: row["Skill Title"]
+      )
+
+      competency.save!
+    end
+  end
+
+  def competency_available?(row)
+    row["Skill Sort Order"] && row["Skill Title"]
```
littleforest, line 59:
```suggestion
    row["Skill Sort Order"].presence && row["Skill Title"].presence
```

Suggest `presence` to make sure a blank string doesn't trigger this to be true.

app/services/import_occupation_standard_work_processes.rb
```diff
@@ -0,0 +1,61 @@
...
+  private
+
+  def create_or_update_competency(row, work_process)
+    Competency.find_or_initialize_by(
+      sort_order: row["Skill Sort Order"],
+      work_process: work_process
+    ).tap do |competency|
+      competency.assign_attributes(
+        title: row["Skill Title"]
+      )
```
littleforest, line 52: Ooo I like this use of `tap` here so as not to have to assign `Competency.find_or_initialize_by` to an external variable and then update it.

spec/factories/data_imports.rb
```diff
@@ -3,5 +3,10 @@
     description { "Here is a description" }
     user
     file { Rack::Test::UploadedFile.new(Rails.root.join("spec", "fixtures", "files", "occupation-standards-template.xlsx"), "application/vnd.openxmlformats-officedocument.spreadsheetml.sheet") }
+
+    factory :data_import_for_hybrid do
+      description { "Here is a description for this Hybrid w/ Max + Min Hours" }
+      file { Rack::Test::UploadedFile.new(Rails.root.join("spec", "fixtures", "files", "comp-occupation-standards-template.xlsx"), "application/vnd.openxmlformats-officedocument.spreadsheetml.sheet") }
```
littleforest, line 9: Should we change the title of this file away from "comp", as it sounds like it is a competency-based template instead of a hybrid example?

littleforest (commented):
Looks good! Just a couple minor suggestions.

spec/factories/data_imports.rb
FerPerales: @littleforest I took the name from Notion, but not sure if those are correct

image

littleforest: Ah okay! We can always rename later if there is confusion.

littleforest (approved):
Looks good to me!

PR #63, opened 2023-02-15
Create Oregon Scraper

jenlu33:
Asana ticket: [https://app.asana.com/0/1203289004376659/1203778503386688/f](https://app.asana.com/0/1203289004376659/1203778503386688/f)

I used this site for scraping: [https://www.oregon.gov/boli/apprenticeship/Pages/apprenticeship-opportunities.aspx](https://www.oregon.gov/boli/apprenticeship/Pages/apprenticeship-opportunities.aspx)


littleforest (commented):
@jenlu33 can you add a spec for this service? See example in #58. I found it easiest to copy the page source into a fixture file and use that.

jenlu33:
I tried filtering the number of jobs shown on the site so that the tests would run faster, but it seems like it still goes through all 84 files. It works, but just takes a while

app/jobs/scraper/oregon_job.rb
```diff
@@ -0,0 +1,48 @@
+class Scraper::OregonJob < ApplicationJob
+  queue_as :default
+
+  def perform
+    base = "https://www.oregon.gov"
+    apprenticeship_url = base + "/boli/apprenticeship/Pages/"
+    browser = Watir::Browser.new
+    browser.goto(apprenticeship_url + "apprenticeship-opportunities.aspx")
+    js_doc = browser.element(css: "tbody").wait_until(&:present?)
+    body = Nokogiri::HTML(js_doc.inner_html)
```
littleforest, line 10: @jenlu33 @FerPerales I think the spec is making the external URL call as far as I can tell? I wonder if the browser allows it to bypass the webmock configuration. I finally had to kill rspec in GH Actions after 4 hours as I think we will need a headless version to work there. Also, the spec alone took 5 minutes locally to run which is too long. 

Could we try stubbing everything? Something like this:

```
js_doc = file_fixture("scraper/some-final-html-after-the-js-has-run.html").read
element = double("element, wait_until: js_doc)
browser_mock = instance_double("Watir::Browser", goto: nil, element: element)
allow(Watir::Browser).to receive(:new).and_return(browser_mock)
```

And then when you copy the HTML file from the website after the JS has run, just remove all but a couple of the table rows so we only have to test that a couple of records are created.

spec/jobs/scraper/oregon_job_spec.rb
```diff
@@ -0,0 +1,40 @@
+require "rails_helper"
+
+RSpec.describe Scraper::OregonJob, type: :job do
+  describe "#perform" do
+    context "when files have not been downloaded previously" do
+      it "downloads pdf files to a standards import record" do
+        stub_responses
```
littleforest, line 7: Insert the 4 lines with the mocking in the above comment here.

FerPerales (approved):
Looks good!

Future @littleforest: we decided to skip the test for now since mocking Watir is harder that expected.

We also discuss about dependency injection for the class, but it's going to involve more effort than my first evaluation. 

Let's discuss a long term solution once you are back but for now, we'll merge and run the scraper in production

PR #68, opened 2023-02-20
Add Apprenticeship.gov bulletin scraper

FerPerales:
Asana ticket: https://app.asana.com/0/1203289004376659/1203948672111227/f

This PR create a new job that fetches the list of bulletins from apprenticeship.gov, then it keeps only the bulletins in docx format that have attachments. Unfortunately, there's no straightway forward to know when a file has attachments and waht you see in the code is the best at the moment.


spec/jobs/scraper/apprenticeship_bulletins_job_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+          described_class.new.perform
+        }.to change(StandardsImport, :count).by(3)
+
+        standard_import = StandardsImport.last
+        expect(standard_import.files.count).to eq 1
+        expect(standard_import.name).to eq "https://www.apprenticeship.gov/sites/default/files/bulletins/Bulletin%202023-50%20New%20NOF%20Appren%20Occ%20Jr%20Cloud%20Engineer.docx"
+        expect(standard_import.organization).to eq "2023-50"
+        expect(standard_import.notes).to eq "From Scraper::AppreticeshipBulletinsJob #{Scraper::AppreticeshipBulletinsJob::BULLETIN_LIST_URL}"
+      end
+    end
```
littleforest, line 18: Could you also add a test that checks the "Attachments. None" skip?

app/jobs/scraper/appreticeship_bulletins_job.rb
```diff
@@ -0,0 +1,49 @@
...
+      end
+    rescue OpenURI::HTTPError
+      Rails.logger.debug "Error while trying to download #{file_name}"
+    end
+  end
+
+  private
+
+  def has_attachments?(docx)
+    docx.paragraphs.none? { |p| p.to_s == "Attachments. None." }
```
littleforest, line 43: I'm a little worried that this is going to be really brittle - were you able to check multiple files to make sure they had this same text?

littleforest (commented):
Can this file be removed? `spec/fixtures/files/bulletins-results.numbers`

app/jobs/scraper/appreticeship_bulletins_job.rb
FerPerales: @littleforest I ran the importer locally and grabbed the expected files. To be 100% sure, I'll need to download and check manually all files (around 1K) :( 

docx is a xml-like file and I couldn't find a more reliable way to check if a file has attachments, but I can do extra research

spec/jobs/scraper/apprenticeship_bulletins_job_spec.rb
FerPerales: @littleforest added

app/jobs/scraper/appreticeship_bulletins_job.rb
littleforest: Let's merge for now, and we can update as this goes into practice.

littleforest (approved):
👍

PR #70, opened 2023-02-20
Fix bug in importing Details tab for data_imports

littleforest:
Organizations have a title instead of a name.
This fleshes out the fixture file to have data
in the "Sponsor Name" column.

Asana ticket: https://app.asana.com/0/1203289004376659/1203289004376679/f


jenlu33 (approved):
Looks good to me!

PR #75, opened 2023-02-24
Add json column to file imports

FerPerales:
Asana ticket: https://app.asana.com/0/1203289004376659/1204028848191911/f

This PR adds a JSONB column to File Import as well as a read-only text area to show its contents

Some FileImport will use this column to store extra information

image


jenlu33 (approved):
Looks good to me!

PR #78, opened 2023-03-07
Add oregon job to scraper task and update to be headless

jenlu33:
I followed this tutorial to make the Oregon scraper headless for production. It looks like I will also need to add some buildpacks in heroku to make sure everything works correctly.

[https://medium.com/@franciscoantunesbarreto1997/configuring-watir-with-ruby-on-rails-for-heroku-e96433854430](https://medium.com/@franciscoantunesbarreto1997/configuring-watir-with-ruby-on-rails-for-heroku-e96433854430)


littleforest (approved):
Nice! Do you have permissions/are you comfortable adding the buildpack on Heroku?

jenlu33:
> Nice! Do you have permissions/are you comfortable adding the buildpack on Heroku?

It looks like I can do it myself! I will let you know if there are any issues

PR #90, opened 2023-03-09
Create New York scraper and spec

jenlu33:
Asana ticket: [https://app.asana.com/0/1203289004376659/1203778503446140/f](https://app.asana.com/0/1203289004376659/1203778503446140/f)

Some rows had an anchor tag with broken links so I had to skip those when saving a StandardsImport


app/jobs/scraper/new_york_job.rb
```diff
@@ -0,0 +1,42 @@
...
+
+  def perform
+    url = "https://dol.ny.gov/apprenticeship/apprenticeship-trades"
+    uri = URI.parse(url)
+    response = Net::HTTP.get_response(uri)
+    html = response.body
+    doc = Nokogiri::HTML(html)
+
+    table_and_nav = doc.css("div.wysiwyg--field-webny-wysiwyg-body")
+    table = table_and_nav.css("tbody")[0]
```
littleforest, line 12: Are there multiple tables on the page? If not, can we just directly get the `table` from the `tbody` rather than starting with the `wysiwyg--...` class? That class seems like that could change.

app/jobs/scraper/new_york_job.rb
```diff
@@ -0,0 +1,42 @@
...
+    table.css("tr").each do |row|
+      next unless row.css("a").present?
+      file_name = row.css("a").first["href"]
+      file_path = ""
+      if /https/.match?(file_name)
+        file_path = file_name.gsub("https://dol.ny.gov/system/files/documents/2022/06", "").gsub(".pdf", "")
```
littleforest, line 18: Can you make `https://dol.ny.gov/system/files/documents/2022/06` a constant, since it is used in two places?

app/jobs/scraper/new_york_job.rb
```diff
@@ -0,0 +1,42 @@
...
+      )
+
+      if standards_import.new_record?
+        begin
+          standards_import.files.attach(
+            io: URI.open("https://www.dol.ny.gov/system/files/documents/2022/06#{file_path}.pdf"),
+            filename: File.basename(file_name)
+          )
+          standards_import.save!
+        rescue
```
littleforest, line 36: What error occurs when the link is broken? It would be better to specifically rescue that error so we don't lose information about other errors that might occur.

littleforest (commented):
@jenlu33 looks good! Just a couple comments.

app/jobs/scraper/new_york_job.rb
jenlu33: There are 3 tables and we only need the first one so I changed it to just grab the first one without querying for the wrapper div first

littleforest (approved):
Looks great!

PR #92, opened 2023-03-13
Move routes to admin namespace

littleforest:
Asana ticket: https://app.asana.com/0/0/1204160879538793/f

Since we will need to have some public facing routes for occupation_standards, this moves all of the existing admin pages under an admin namespace.


FerPerales (approved):
Thanks for tackling this!

PR #99, opened 2023-03-16
Onet code in rake task

fatimaburke:
Asana ticket:


littleforest (commented):
@fatimaburke were you still running into issues with the rake task? I don't think we necessarily needs these variable re-naming changes.

fatimaburke:
Hey! @littleforest these changes were made as a result of a working session with just to get me unblocked. Maybe @Pcushing can speak to the rationale of the naming conventions, if any but it could be just a duplicate.

littleforest:
I'm actually going to close, as I'm in the process of moving the bulk of these tasks to a service with some specs so we don't run into this same issue again.

PR #106, opened 2023-03-21
Scrape washington site

jenlu33:
Asana ticket: [https://app.asana.com/0/1203289004376659/1203778503386686/f](https://app.asana.com/0/1203289004376659/1203778503386686/f)

I used the watir gem in this scraper just like the Oregon job, but had the same issues with the spec so it is ignored for now.


app/jobs/scraper/washington_job.rb
```diff
@@ -0,0 +1,55 @@
...
+    program_ids = []
+
+    loop do
+      find_table = browser.element(css: "tbody").wait_until(&:present?)
+
+      begin
+        table = Nokogiri::HTML(find_table.inner_html)
+      rescue Watir::Wait::TimeoutError
+        sleep 60
+        retry
```
littleforest, line 22: @jenlu33 is this timeout happening frequently? I wonder if we should only retry for X number of times so the job potentially doesn't get stuck forever. Something like this:

```
      begin
        retries ||= 0
        table = Nokogiri::HTML(find_table.inner_html)
      rescue Watir::Wait::TimeoutError
        sleep 60
        if (retries += 1) <= MAX_RETRIES
          retry
        else
          # Raise error or do nothing 
        end
      end
```

jenlu33: It doesn't happen very often, I'll add that in

app/jobs/scraper/washington_job.rb
```diff
@@ -0,0 +1,56 @@
...
+    loop do
+      find_table = browser.element(css: "tbody").wait_until(&:present?)
+      begin
+        attempts ||= 0
+        table = Nokogiri::HTML(find_table.inner_html)
+      rescue Watir::Wait::TimeoutError
+        attempts += 1
+        sleep 60
+        retry if attempts <= 5
+      end
```
littleforest, line 24: Sorry I didn't catch this before -- now with the retry added, you will need to wrap all of the lines (26-33) below also in the `begin` block, otherwise, if this times out 5 times, then it will just continue on to line 26, but there will be no `table` object and it will error.

littleforest: And probably it would be good to report an error if it does timeout, so at least we are alerted if the script doesn't run.

```
rescue Watir::Wait::TimeoutError => e
  attempts += 1
  if attempts <= 5
    sleep 60
    retry
  else
    Rails.error.report(e)
  end
end
```

littleforest (approved):
🎉

PR #111, opened 2023-04-03
Home page structure

fatimaburke:
Home page Design

Screen Shot 2023-03-30 at 9 13 22 PM
Screen Shot 2023-03-30 at 9 13 53 PM


fatimaburke:
@littleforest Can you look into why the rspec test is failing?

PR #113, opened 2023-04-03
Admin can view and edit all attributes for occupation standards

FerPerales:
Asana ticket:https://app.asana.com/0/1203289004376659/1203989413049377/f

Show view:

image
image

Edit view:

image
image


littleforest (approved):
This looks good! Going forward, I don't think we need to worry too much about system specs for the admin pages unless we are trying to test custom behavior. We can trust that Administrate works as designed.

PR #117, opened 2023-04-06
Display real data on public occupation standards page

littleforest:
Asana ticket: https://app.asana.com/0/1203289004376659/1204349715012993/f


littleforest:
Looks great, thanks @fatimaburke!

PR #122, opened 2023-04-12
Add ability to delete source file on admin

jenlu33:
Asana ticket: [https://app.asana.com/0/1203289004376659/1203989413049379/f](https://app.asana.com/0/1203289004376659/1203989413049379/f)

I'm not really sure how Administrate works, but it looks like it automatically added the "Destroy" button for me when I added it into the routes? And then there is the popup to confirm the delete also.

Screenshot 2023-04-12 at 12 03 21 PM

Screenshot 2023-04-12 at 11 44 26 AM

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -195,6 +195,7 @@ GEM
       activerecord (>= 4.1)
       concurrent-ruby
       railties (>= 4.1)
+    jwt (2.7.0)
     kaminari (1.2.2)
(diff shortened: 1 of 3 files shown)


littleforest (approved):
Perfect! Going forward, we don't need to write new tests for the Administrate CRUD routes; we can assume that they are tested already as part of the gem. We should just add tests for custom routes or behaviors that we add.

PR #124, opened 2023-04-17
Add URL link in source file show view

FerPerales:
Asana ticket: https://app.asana.com/0/1203289004376659/1204327785987146/f

image


jenlu33 (approved):

littleforest:
@FerPerales sorry I am late to the review here -- is the link clickable? If not, can we make it so?

PR #136, opened 2023-04-24
Update individual Occupation Standard view

fatimaburke:
Asana ticket: https://app.asana.com/0/1203289004376659/1204443182031652/f

image


app/views/occupation_standards/show.html.erb
```diff
@@ -16,111 +16,228 @@
...
+                <div class="w-full h-full rounded py-4 px-2 mr-4 mb-4 border-2 border-seafoam-200 bg-white">
+                    <div class="text-primary-600 font-bold">
+                        Mechatronics Technician
+                    </div>
+                    <div class="text-orange-800">
+                        <span class="text-yonder-500">Hybrid</span>
+                        <span>&#x2022;</span>
+                        <span>17-3024.00</span>
+                    </div>
+                </div>
```
littleforest, line 207: @fatimaburke can you wrap these in `a` tags with just `href="#"` for now so that they could be clicked to go view that standard?

littleforest: @fatimaburke I was thinking more that each box would be clickable and could take you to view that individual standard. 

Screen Shot 2023-04-25 at 8 54 15 AM

fatimaburke: @littleforest Thanks for the clarity. I'll update.

littleforest: @fatimaburke thanks for the update! Looks great. For some reason now the 2nd box has the blue color for the ONET code. I cannot figure it out, the blocks all look the same? Do you know what is causing that?

Screen Shot 2023-04-25 at 11 05 02 AM

littleforest (approved):
Looks great! 🎉

PR #137, opened 2023-04-25
Converter users only see Source files in Administrate

littleforest:
Asana ticket: https://app.asana.com/0/1203289004376659/1204452198889951/f

This sets up Pundit to authorize the converter user to only access the Source Files and Data Import records.

Screen Shot 2023-04-25 at 10 16 31 AM
Screen Shot 2023-04-25 at 10 16 38 AM

Screen Shot 2023-04-25 at 10 16 53 AM
Screen Shot 2023-04-25 at 10 18 16 AM


spec/factories/users.rb
```diff
@@ -1,5 +1,7 @@
 FactoryBot.define do
   factory :user do
+    traits_for_enum :role, User.roles
```
FerPerales, line 3: TIL: traits for enum. That's clever!

FerPerales (approved):
Looks good! Wondering if we should add unit tests for the policies or only rely on request specs for that 🤔

littleforest:
> Wondering if we should add unit tests for the policies or only rely on request specs for that

Do you usually write tests for the policies? I usually just use request specs, but probably writing tests for the policies would be better, and then I could skip the more verbose request specs?

FerPerales:
> > Wondering if we should add unit tests for the policies or only rely on request specs for that
> 
> Do you usually write tests for the policies? I usually just use request specs, but probably writing tests for the policies would be better, and then I could skip the more verbose request specs?

Depends on the project. For this one, it seems our policies will be relatively straight forward, but for another project that has more complex logic, I do

littleforest:
@FerPerales thanks for the policy spec suggestion! I made that addition.

FerPerales (approved):

PR #142, opened 2023-04-27
Allow user to search for occupation standard by title

FerPerales:
Asana ticket: https://app.asana.com/0/1203289004376659/1204443316982593/f

https://user-images.githubusercontent.com/1335038/234978852-e9ed8788-5dce-486a-ac5a-231ddedbdd5f.mp4

### DIFF
--- a/app/controllers/occupation_standards_controller.rb
+++ b/app/controllers/occupation_standards_controller.rb
@@ -1,6 +1,6 @@
 class OccupationStandardsController < ApplicationController
   def index
-    @occupation_standards_search = OccupationStandardQuery::Container.new(search_term: search_term_params)
+    @occupation_standards_search = OccupationStandardQuery::Container.new(q: params[:q])
 
     @occupation_standards = OccupationStandardQuery.run(
       OccupationStandard.includes(:organization, occupation: :onet),
@@ -18,7 +18,7 @@ class OccupationStandardsController < ApplicationController
 
   def search_term_params
     {
-      search_term: params[:q]
+      q: params[:q]
     }
   end
 end
--- a/app/controllers/occupation_standards_controller.rb
+++ b/app/controllers/occupation_standards_controller.rb
@@ -1,11 +1,24 @@
 class OccupationStandardsController < ApplicationController
   def index
-    @pagy, @occupation_standards = pagy(
-      OccupationStandard.includes(:organization, occupation: :onet)
+    @occupation_standards_search = OccupationStandardQuery::Container.new(search_term: search_term_params)
+
+    @occupation_standards = OccupationStandardQuery.run(
+      OccupationStandard.includes(:organization, occupation: :onet),
+      search_term_params
     )
+
+    @pagy, @occupation_standards = pagy(@occupation_standards)
   end
 
   def show
     @occupation_standard = OccupationStandard.find(params[:id])
+  end
+
+  private
+
+  def search_term_params
+    {
+      search_term: params[:q]
+    }
   end
 end
--- a/app/controllers/occupation_standards_controller.rb
+++ b/app/controllers/occupation_standards_controller.rb
@@ -1,6 +1,6 @@
 class OccupationStandardsController < ApplicationController
   def index
-    @occupation_standards_search = OccupationStandardQuery::Container.new(q: params[:q])
+    @occupation_standards_search = OccupationStandardQuery::Container.new(search_term_params: search_term_params)
 
     @occupation_standards = OccupationStandardQuery.run(
       OccupationStandard.includes(:organization, occupation: :onet),
--- a/app/queries/occupation_standard_query.rb
+++ b/app/queries/occupation_standard_query.rb
@@ -1,6 +1,6 @@
 class OccupationStandardQuery
   ATTRIBUTES = %i[
-    search_term
+    q
   ]
 
   # Container is a wrapper for the params involved in the search
@@ -11,6 +11,6 @@ class OccupationStandardQuery
     return occupation_standards if search_term_params.blank?
 
     occupation_standards
-      .by_title(search_term_params[:search_term])
+      .by_title(search_term_params[:q])
   end
 end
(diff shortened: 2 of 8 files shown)


spec/system/occupation_standards/index_spec.rb
```diff
@@ -10,4 +10,20 @@
...
+    mechanic = create(:occupation_standard, title: "Mechanic")
+    pipe_fitter = create(:occupation_standard, title: "Pipe Fitter")
+
+    visit occupation_standards_path
+
+    fill_in "q", with: "Mechanic"
+
+    click_button "button"
+
+    expect(page).to have_text "Showing Results for Mec"
```
littleforest, line 24: Any reason not to put the whole string here?
```suggestion
    expect(page).to have_text "Showing Results for Mechanic"
```

app/queries/occupation_standard_query.rb
```diff
@@ -0,0 +1,16 @@
+class OccupationStandardQuery
+  ATTRIBUTES = %i[
+    search_term
+  ]
+
+  # Container is a wrapper for the params involved in the search
+
+  Container = Struct.new(*ATTRIBUTES, keyword_init: true)
```
littleforest, line 8: What is `keyword_init` for?

spec/models/occupation_standard_spec.rb
```diff
@@ -83,4 +83,20 @@
...
+
+      expect(described_class.by_title("A").pluck(:id)).to match_array([first_occupation.id])
+    end
+
+    it "returns all records if title not provided" do
+      first_occupation = create(:occupation_standard, title: "AAAAAA")
+      second_occupation = create(:occupation_standard, title: "ZZZZZZ")
+
+      expect(described_class.by_title("").pluck(:id)).to match_array([first_occupation.id, second_occupation.id])
+    end
```
littleforest, line 100: Could we add another spec or two that handles two word titles, like "pipe fitter"?

littleforest (commented):
Looks good! Just a couple minor comments.

app/queries/occupation_standard_query.rb
FerPerales: @littleforest this allows initialize an struct with a list of keywords

https://www.bigbinary.com/blog/ruby-2-5-allows-creating-structs-with-keyword-arguments

littleforest (commented):
In the URL after I enter a search, I'm getting a "button" term in there -- can that be removed?

http://localhost:3000/occupation_standards?q=pipe&button=

app/controllers/occupation_standards_controller.rb
```diff
@@ -1,11 +1,24 @@
...
 
   def show
     @occupation_standard = OccupationStandard.find(params[:id])
   end
+
+  private
+
+  def search_term_params
+    {
+      search_term: params[:q]
```
littleforest, line 21: What do you think about using `q` here as the key also? I'm kind of torn. I like `search_term` because it is a well-named variable, but I feel like the `search_term[:search_term]` for the Struct gets a little confusing.

FerPerales: @littleforest makes sense. Updated

FerPerales:
> In the URL after I enter a search, I'm getting a "button" term in there -- can that be removed?
> 
> http://localhost:3000/occupation_standards?q=pipe&button=

Removed

app/views/occupation_standards/index.html.erb
```diff
@@ -23,11 +23,11 @@
 
 <section aria-labelledby="occupation-title" class="bg-seafoam-50 dark:bg-gray-900">
     <div class="mx-auto max-w-screen-xl py-8 px-4 sm:py-16 lg:px-6">
-        <% if @occupation_standards_search.search_term[:search_term].present? %>
+        <% if @occupation_standards_search.q.present? %>
```
littleforest, line 26: Now I feel like I'm confused. I thought this would look like: `@occupation_standards_search.search_term[:q].present?`

app/controllers/occupation_standards_controller.rb
```diff
@@ -1,6 +1,6 @@
 class OccupationStandardsController < ApplicationController
   def index
-    @occupation_standards_search = OccupationStandardQuery::Container.new(search_term: search_term_params)
+    @occupation_standards_search = OccupationStandardQuery::Container.new(q: params[:q])
```
littleforest, line 3: Why not keep using `search_term_params`?

littleforest: Oh, I think I see the confusion. Could we do:

```
search_term: search_term_params
```

like before, but have `search_term_params = { q: q }` ?

app/views/occupation_standards/index.html.erb
littleforest: That way we can have `q` as a general search query term, but we can also start adding `title`, `onet`, etc. as more specific filters?

FerPerales: Search term is the query from the UI. Nesting q inside search_term may not make a lot of sense since we won't have multiple search terms

littleforest: What about adding filters? Won't we use the same Query interface?

FerPerales: We'll use the same, but not sure if we have defined how we are going to implement the filters. We can use q for title, and then specific fields for the filters 🤔

littleforest: My vote is either:

```
@occupation_standards_search = OccupationStandardQuery::Container.new(search_term: search_term_params)

def search_term_params
 {q: params[:q]}
end
```

or

```
@occupation_standards_search = OccupationStandardQuery::Container.new(search_term: params[:q])
```

But I think my least favorite is just `q` running around by itself somewhat ill-defined.

littleforest: And I think my vote is really the top one, so that we can pass other parameters for filters (thinking ahead to filtering by the `national_standard_type` field by itself).

PR #146, opened 2023-05-01
Hook up search on Home page

littleforest:
Asana ticket: https://app.asana.com/0/1203289004376659/1204452198889953/f

Needs some styling help:

Before:

Screen Shot 2023-05-01 at 9 53 35 AM

Now:

Screen Shot 2023-05-01 at 9 57 40 AM


littleforest:
@fatimaburke it's now off-center on the `/occupation_standards` page:

Screen Shot 2023-05-01 at 10 38 55 AM

Is there a way to put the styling just into the divs wrapping the form so that they form can be shared?

PR #148, opened 2023-05-01
Account for no rapids code or organization

jenlu33:
Asana ticket: [https://app.asana.com/0/1203289004376659/1204504222387378/f](https://app.asana.com/0/1203289004376659/1204504222387378/f)

It looks like some occupation standards do not have a rapids code so I changed it so that the gsub only happens if there is one.
Screenshot 2023-05-01 at 4 08 05 PM

Same with organization, it looks like some related instructions do not have an organization so it starts off as nil.
Screenshot 2023-05-01 at 4 11 26 PM


littleforest (commented):
@jenlu33 can you update the spec file and add rows that account for these changes?

spec/services/import_occupation_standard_details_spec.rb
```diff
@@ -85,8 +85,8 @@
         ca = create(:state, abbreviation: "CA")
         ca_oa = create(:registration_agency, state: ca, agency_type: :oa)
 
-        create(:onet, code: "13-1071.01")
-        occupation = create(:occupation, rapids_code: "0157")
+        onet = create(:onet, code: "13-1071.01")
+        occupation = create(:occupation, onet: onet)
```
littleforest, line 89: @jenlu33 could you add an additional line in the spec file and check for a new record getting save? -- I want to keep this check for "0157" so that we have a case that accounts for rapid codes that start with a 0.

jenlu33: @littleforest What would be the best way to do this? I noticed that `ImportOccupationStandardDetails` only takes in one row

littleforest (approved):
Looks great, thanks for adding the extra spec.

PR #150, opened 2023-05-02
Add Filters to Search Form

fatimaburke:
Asana ticket: https://app.asana.com/0/1203289004376659/1204452198889977/f

Screen Shot 2023-05-02 at 4 05 06 PM
Screen Shot 2023-05-02 at 4 05 20 PM


littleforest (commented):
@fatimaburke the filters don't seem to have the same padding that they did on the Figma designs. Just wanted to see if this was correct? Do I need to do anything special to get the assets to work?

Screen Shot 2023-05-02 at 2 24 32 PM

littleforest:
@fatimaburke could you also merge in main and fix merge conflicts?

config/tailwind.config.js
```diff
@@ -7,7 +7,8 @@ module.exports = {
     './public/*.html',
     './app/helpers/**/*.rb',
     './app/javascript/**/*.js',
-    './app/views/**/*.{erb,haml,html,slim}'
+    './app/views/**/*.{erb,haml,html,slim}',
+    './node_modules/flowbite/**/*.js',
```
littleforest, line 11: Oh shoot, I totally removed `node_modues` -- @FerPerales is there no way to get flowbite without this?

FerPerales: @littleforest based on the research we did in #20 , this is the recommended way to do it. We were reluctant to introduce the dependency only for closing flash messages, but I think @fatimaburke is considering using more elements

littleforest: @FerPerales but if we are loading flowbite through importmaps, can we access it through there instead?

FerPerales: @littleforest what we are doing through importmaps is adding turbo support, according to the [docs](https://flowbite.com/docs/getting-started/rails/#importmap). We can remove it and see if it works

littleforest (approved):
🎉 looks great! Would like Fer to review too before merging.

PR #154, opened 2023-05-03
Search results can be filtered by rapids code

littleforest:
Asana ticket: https://app.asana.com/0/1203289004376659/1204452198889965/f

Allow searching by RAPIDS code for the main search box. This also modifies the import script to store the rapids code of the occupation directly on the occupation_standard record if the "RAPIDS code" field is blank on the import script.

Screen Shot 2023-05-03 at 1 49 36 PM


FerPerales (approved):
:shipit:

PR #157, opened 2023-05-04
Allow searching on state

littleforest:
Asana ticket: https://app.asana.com/0/1203289004376659/1204452198889961/f

This changes the State field in the filters section to be a drop-down that lists all the states, and adds filtering by state.

Screen Shot 2023-05-04 at 11 30 12 AM

### DIFF
--- a/app/helpers/occupation_standards_helper.rb
+++ b/app/helpers/occupation_standards_helper.rb
@@ -8,4 +8,8 @@ module OccupationStandardsHelper
       "hidden"
     end
   end
+
+  def filters_aria_expanded
+    filters_class == "hidden" ? "false" : "true"
+  end
 end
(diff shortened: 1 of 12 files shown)


app/helpers/occupation_standards_helper.rb
```diff
@@ -2,4 +2,10 @@ module OccupationStandardsHelper
   def ojt_type_display(occupation_standard)
     occupation_standard.ojt_type&.titleize
   end
+
+  def filters_class
+    unless params[:state_id].present?
+      "hidden"
+    end
+  end
```
littleforest, line 10: I think we should probably put all the filter params within a `filter:` hash, since adding each individual filter here is going to be somewhat annoying. But I figured it could wait until the next filter that we add.

FerPerales (approved):

PR #163, opened 2023-05-05
Hide View Redacted Document

FerPerales:
Asana ticket: https://app.asana.com/0/1203289004376659/1204504222387386/f

If occupation standard was created from a public document:

image

If it wasn't:

image

### DIFF
--- a/app/views/occupation_standards/show.html.erb
+++ b/app/views/occupation_standards/show.html.erb
@@ -77,22 +77,25 @@
                     <div class="w-full space-y-6">
                         <div class="text-center text-lg font-bold text-slate-300">Documents</div>
                         <% if @occupation_standard.public_document? %>
-                            <%= link_to @occupation_standard.original_file_url do %>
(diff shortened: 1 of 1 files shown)


app/views/occupation_standards/show.html.erb
```diff
@@ -77,22 +77,25 @@
...
+
+                              <div class="font-bold">
+                                View Original Document
+                              </div>
+                            </div>
+                          <% end %>
+                        <% else %>
+                          <p class="text-white">No documents available at the moment</p>
                         <% end %>
+                        <!--
```
littleforest, line 98: Do we want to comment this out, or should this be `unless public_document` ...?

FerPerales: @littleforest we are going to enable this once we have a way to redact documents. This will chance one we have that feature enabled

littleforest (approved):

PR #164, opened 2023-05-05
Enable assets pipeline fallback in production

FerPerales:
Asana ticket:

This may fix the missing assets in production. It's not the best long-term solution, though


littleforest (approved):
Could you make an Asana ticket to follow up on this later?

PR #170, opened 2023-05-05
Find WorkProcess only by title

FerPerales:
Asana ticket: https://app.asana.com/0/1203289004376659/1204532764328547/f

Make job find Work Processes only by title and not by sort_order, which is unique and was causing to create duplicated work processes with the same name


littleforest (approved):
🚢

PR #172, opened 2023-05-05
Add competencies list to work processes accordion

FerPerales:
Asana ticket:

image

Display competencies under each Work Process as a list


littleforest (approved):
:shipit:

PR #182, opened 2023-05-10
Remove HTML tags from competency title

FerPerales:
Asana ticket: https://app.asana.com/0/1203289004376659/1204546181205105/f

image


littleforest (approved):
What do you think about adding a quick spec?

PR #187, opened 2023-05-11
Extend ActiveStorage service URL expiration

FerPerales:
Asana ticket: https://app.asana.com/0/1203289004376659/1204546181205096/f

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -37,5 +37,7 @@ module ApprenticeshipStandardsDotOrg
     config.after_initialize do
       Rails.error.subscribe(ErrorSubscriber.new)
     end
+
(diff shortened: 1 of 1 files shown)


config/application.rb
```diff
@@ -37,5 +37,7 @@ class Application < Rails::Application
     config.after_initialize do
       Rails.error.subscribe(ErrorSubscriber.new)
     end
+
+    config.active_storage.service_urls_expire_in = 15.minutes
```
littleforest, line 41: I wonder if we should make it longer? Also, what do you think about implementing a more permanent solution that refreshes the URL?

littleforest (approved):

PR #192, opened 2023-05-15
Improve setup process

sarahraqueld:
Run the seeds task as part of the setup script. This will help new teammates joining the project as they will already have an user created and will be able to login on the app directly after setting up. 

Add a yarn version to .tool-versions. This will prevent future teammates who use `asdf` from running the same problem as a did when running the setup script: `No version is set for yarn`

Add the `x86_64-darwin-22` platform to Gemfile.lock. This was an automatic addition after setting up the project.


littleforest (approved):
nice, thanks!

FerPerales (approved):

PR #193, opened 2023-05-15
Add "Needs Support" status to SourceFile and make it searchable

littleforest:
Asana ticket: https://app.asana.com/0/1203289004376659/1204601802991338/f

Screen Shot 2023-05-15 at 2 06 42 PM
Screen Shot 2023-05-15 at 2 06 54 PM

### DIFF
--- a/app/controllers/admin/source_files_controller.rb
+++ b/app/controllers/admin/source_files_controller.rb
@@ -37,13 +37,17 @@ module Administrate
     end
 
     def query_values
-      array = super
-      term = array.first
-      array + [term, SourceFile.statuses[term.parameterize(separator: "_")]]
+      values = super
+      term = values.first
+      values + [term, db_value_for_status(term)]
     end
 
     def search_results(resources)
       super.left_joins(active_storage_attachment: :blob)
+    end
+
+    def db_value_for_status(term)
+      SourceFile.statuses[term.parameterize(separator: "_")]
     end
   end
 end
--- a/app/controllers/admin/source_files_controller.rb
+++ b/app/controllers/admin/source_files_controller.rb
@@ -32,12 +32,14 @@ module Administrate
       super
         .split(" OR ")
         .push("LOWER(active_storage_blobs.filename) LIKE ?")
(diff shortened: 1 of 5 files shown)


app/controllers/admin/source_files_controller.rb
```diff
@@ -32,12 +32,14 @@ def query_template
...
         .push("LOWER(active_storage_blobs.filename) LIKE ?")
+        .push("status = ?")
         .join(" OR ")
     end
 
     def query_values
       array = super
-      array + [array.first]
+      term = array.first
+      array + [term, SourceFile.statuses[term.parameterize(separator: "_")]]
```
sarahraqueld, line 42: Do you know if this part of the app is often read/changed? 

I ask because I think this line is a bit complex, and in case we do read it often, I think it would be valuable to maybe extract to more simpler/atomic lines to make debugging/following along the code execution easier. If not, then it's fine!

sarahraqueld (approved):
Don't have the overall product understanding yet, but code-wise, looks good! Left a comment :)

app/controllers/admin/source_files_controller.rb
littleforest: It is calling `super` on [this Administrate method](https://github.com/thoughtbot/administrate/blob/44ad6ce1072fbd897f1faaf7231b3b5a0b92f535/lib/administrate/search.rb#L103), so `array` just looks like: `array = ["%searchterm%", "%searchterm", ..., "%searchterm"]`. Do you think something like:

```
array + [term, db_value_for_status]
```

would be better, where I add a method called `db_value_for_status`? And maybe use "values" instead of "array"?

```
def query_values
  values = super
  term = values.first
  values + [term, db_value_for_status(term)]
end

private

def db_value_for_status(term)
  SourceFile.statuses[term.parameterize(separator: "_")
end
```

Or let me know if you had something else in mind.

sarahraqueld: Yes, I like that! Thanks 😊

PR #194, opened 2023-05-15
[WIP] Export Occupation Standards as .docx

sarahraqueld:
Asana ticket: https://app.asana.com/0/1203289004376659/1204546181205130/f

## Why?

Users would like a working document with the occupation standards data.

## How?

First, Add the Sablon gem - A word document templating tool. This specific gem was chosen as it's the same used on another project recommended by Jeanine. It seems like it is maintained and it is easy to use. 

Then, customize the existing occupation standard show action to add a new option of return format.¹
Once this format is requested, the app will generate a .docx document to be downloaded by our users.

To make reviewing easier, this work is broken in *at least* two PRs, here, I introduce the routes, controller and document generation, and in a next PR, I'll work on improving the docx template.

## Screenshots

Screenshot 2023-05-17 at 11 23 55

Sample of a document generated 

Screenshot 2023-05-16 at 15 49 02

Button to Download a Working Copy added to the Occupation Standards show page
 
----

¹: I had considered adding a new controller or new controller action to deal with this export, but after checking this [rails documentation page](https://guides.rubyonrails.org/action_controller_overview.html#restful-downloads), it makes sense that this .docx file "can be considered just another representation of the client resource".


sarahraqueld:
@FerPerales told me know that this app is still not being used in production, so I think it is OK that this is reviewed and merged before it is ready. Let me know if this is incorrect and I should only merge once the template is done.

FerPerales:
@sarahraqueld if the feature works and the only missing part is the final template, we can probably tackle that in another task. @littleforest what are your thoughts?

app/models/occupation_standard_export.rb
```diff
@@ -0,0 +1,28 @@
...
+  end
+
+  def filename
+    title = occupation_standard.title.parameterize
+    "#{title}-#{Time.zone.now.iso8601}.docx"
+  end
+
+  private
+
+  def context_hash
```
sarahraqueld, line 23: This method will grow as we'll need to add more data into the docx.

littleforest:
Agreed that we can merge in sections as needed since app is not live yet.

FerPerales (approved):

app/models/occupation_standard_export.rb
```diff
@@ -0,0 +1,29 @@
...
+    {
+      occupation_standard: occupation_standard,
+      ojt_type: occupation_standard.ojt_type&.titleize
```
littleforest, line 26: There is a `ojt_type_display` method in the OccupationStandardsHelper module which is identical -- should we just make this a method in the `OccupationStandard` model and call it from there instead?

sarahraqueld:
I talked to Patrick and he let me know that there are a few people that use the site and I'd not be a good idea to have unfinished work live, specially related to documents. So I'll make the download button hidden for now

app/models/occupation_standard_export.rb
sarahraqueld: Yes, agreed. Done. I will only introduce the method and use on this context_hash, and will open a follow up PR with the changes to the places that use the ojt_type_display helper.

PR #197, opened 2023-05-16
Return up to five similar programs

FerPerales:
Asana ticket: https://app.asana.com/0/1203289004376659/1204546181205088/f

image

### DIFF
--- a/app/models/occupation_standard.rb
+++ b/app/models/occupation_standard.rb
@@ -19,7 +19,7 @@ class OccupationStandard < ApplicationRecord
   validates :title, :ojt_type, presence: true
   validates :registration_agency, presence: true, unless: :national?
 
-  MAX_SIMILAR_PROGRAM_TO_DISPLAY = 5
+  MAX_SIMILAR_PROGRAMS_TO_DISPLAY = 5
 
   scope :by_title, ->(title) do
     if title.present?
@@ -110,7 +110,7 @@ class OccupationStandard < ApplicationRecord
       "title ILIKE ?", "%#{self.class.sanitize_sql_like(title).split.join("%")}%"
     ).where.not(
       id: id
-    ).limit(MAX_SIMILAR_PROGRAM_TO_DISPLAY)
+    ).limit(MAX_SIMILAR_PROGRAMS_TO_DISPLAY)
   end
 
   private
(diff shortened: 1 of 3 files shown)


app/views/occupation_standards/show.html.erb
```diff
@@ -225,23 +225,23 @@
...
-                                <span class="text-sky-800"><%= ojt_type_display(similar_program) %></span>
-                                <span>&#x2022;</span>
-                                <span><%= similar_program.onet_code %></span>
-                            </div>
-                        <% end %>
-                    <% end %>
-                </div>
+          <div class="similar-programs">
+            <h3 class="font-bold text-2xl text-primary-600 pt-4">Similar Programs</h3>
+            <div class="py-4 grid-cols-5">
```
fatimaburke, line 230: So that these could expand to full width on mobile, I'd replace `grid-cols-5` to `flex flex-col lg:flex-row`

app/views/occupation_standards/show.html.erb
```diff
@@ -225,23 +225,23 @@
...
+              <% @occupation_standard.similar_programs.each do |similar_program| %>
+                <%= link_to similar_program, class: "inline-block rounded py-4 px-2 mr-4 mb-4 border-2 border-seafoam-200 bg-white" do %>
```
fatimaburke, line 232: and likewise replace `inline-block` with `lg:w-1/5`

app/models/occupation_standard.rb
```diff
@@ -104,7 +104,11 @@ def related_instructions_hours
   end
 
   def similar_programs
-    OccupationStandard.where("title ILIKE ?", "%#{self.class.sanitize_sql_like(title).split.join("%")}%") - [self]
+    OccupationStandard.where(
+      "title ILIKE ?", "%#{self.class.sanitize_sql_like(title).split.join("%")}%"
+    ).where.not(
+      id: id
+    ).limit(5)
```
littleforest, line 111: Can we make this a constant?

spec/models/occupation_standard_spec.rb
```diff
@@ -316,6 +316,13 @@
       expect(occupation_standard.similar_programs.pluck(:id)).to match_array [similar_program1.id, similar_program2.id]
     end
 
+    it "returns up to 5 occupations" do
+      occupation_standard = create(:occupation_standard, title: "Human Resource Specialist")
+      create_list(:occupation_standard, 10, title: "Human Resource Specialist")
```
littleforest, line 321: If we make the `5` a constant, then can you stub it to be something slightly smaller so we are creating fewer DB records in the test?

fatimaburke:
Looks good to me on the visual side!

littleforest (approved):
🚢

PR #198, opened 2023-05-16
Allow admins to upload a redacted document

FerPerales:
Asana ticket: https://app.asana.com/0/1203289004376659/1204614576547712/f

Admins can now upload a redacted version of a document from the Admin Occupation Standard edit view

image

When image uploaded, they can see a preview:

image

Show view when there is no redacted document:

image

User can download the document when available:

image


app/views/occupation_standards/show.html.erb
```diff
@@ -114,19 +128,6 @@
                         <% else %>
                           <p class="text-white">No documents available at the moment</p>
                         <% end %>
-                        <!--
-                        <button class="flex w-full items-center space-x-4 rounded bg-slate-700 p-2 text-left text-white hover:bg-slate-600">
```
sarahraqueld, line 118: Looking at my [task](https://app.asana.com/0/1203289004376659/1204546181205130/f) and the figma files, I assumed this button would change its label from "View Redacted Document" to "Download Working Copy", mostly because of this comment from Patrick:

"The "Download Working Copy" link is hidden on the live site at the moment but you can see it in the figma designs".

Since there isn't anything labeled "Download Working Copy" on the codebase, I assumed this button was being renamed. From your PR, it seems like this is not the case, and I should add a new button. Does it seem right to you?

FerPerales: @sarahraqueld yes. There are three different documents and "Download Working Copy" is a brand-new one independent from "View Redacted Document" and "View Original Document"

app/views/occupation_standards/show.html.erb
```diff
@@ -95,7 +95,21 @@
                 <div class="flex h-full w-full md:w-1/3 items-center rounded bg-slate-700 p-4"> <!-- documents - done -->
                     <div class="w-full space-y-6">
                         <div class="text-center text-lg font-bold text-slate-300">Documents</div>
-                        <% if @occupation_standard.public_document? %>
+                        <% if @occupation_standard.redacted_document.present? %>
```
littleforest, line 98:
```suggestion
                        <% if @occupation_standard.redacted_document.attached? %>              
```

https://api.rubyonrails.org/classes/ActiveStorage/Attached/One.html#method-i-attached-3F

spec/system/occupation_standards/show_spec.rb
```diff
@@ -42,4 +42,35 @@
...
+    expect(page).to have_no_text "View Redacted Document"
+  end
+
+  it "shows message if neither public nor redacted document are available" do
+    occupation_standard = create(:occupation_standard, :with_data_import)
+    allow_any_instance_of(OccupationStandard).to receive(:public_document?).and_return(false)
+
+    visit occupation_standard_path(occupation_standard)
+
+    expect(page).to have_text "No documents available at the moment"
```
littleforest, line 72: Can we say "at this time" instead of "at the moment"? I feel like "at the moment" is a little casual and too short-term.

app/views/occupation_standards/show.html.erb
```diff
@@ -114,21 +128,21 @@
                             </div>
                           <% end %>
                         <% else %>
-                          <p class="text-white">No documents available at the moment</p>
+                          <p class="text-white">No documents available at this time</p>
                         <% end %>
-                        <%= link_to occupation_standard_path(@occupation_standard, format: :docx), class: "hidden" do %>
```
sarahraqueld, line 119: @FerPerales since I merged my PR first, seems like there's a conflict here, as we don't actually want to revert this change. Will you rebase and undo these removals? Thanks!

PR #199, opened 2023-05-16
Style Active Navigation Item

fatimaburke:
Asana ticket: https://app.asana.com/0/1203289004376659/1204607522639581/f

Screen Shot 2023-05-16 at 6 08 07 PM

### DIFF
--- a/app/views/layouts/_nav.html.erb
+++ b/app/views/layouts/_nav.html.erb
@@ -16,7 +16,7 @@
                 </li>
                 -->
                 <%= content_tag(:li, link_to("Occupations", occupation_standards_path, class: class_names("block py-3 px-4 text-white"), "aria-current": "page"), class: class_names("block lg:inline hover:shadow-[0_3px_0_0_rgba(249,111,0,1)]", {active: current_page?(occupation_standards_path)})) %>
-                <%= content_tag(:li, link_to("Sate", state_standards_path, class: class_names("block py-3 px-4 text-white"), "aria-current": "page"), class: class_names("block lg:inline hover:shadow-[0_3px_0_0_rgba(249,111,0,1)]", {active: current_page?(state_standards_path)})) %>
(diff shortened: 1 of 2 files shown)


app/views/layouts/_nav.html.erb
```diff
@@ -9,29 +9,19 @@
...
+                <%= content_tag(:li, link_to("National", national_standards_path, class: class_names("block py-3 px-4 text-white"), "aria-current": "page"), class: class_names("block lg:inline hover:shadow-[0_3px_0_0_rgba(249,111,0,1)]", {active: current_page?(national_standards_path)})) %>
                 <li>
                   <span class="block w-0 h-0 md:h-6 md:w-px bg-seafoam-400"></span>
                 </li>
-                <li class="block border-t border-1 border-seafoam-400 md:border-none lg:inline hover:font-bold hover:shadow-[0_3px_0_0_rgba(249,111,0,1)]">
-                  <%= link_to "Get Involved", home_page_path, class: "block py-3 px-4 text-white hover:shadow-[0_3px_0_0_rgba(249,111,0,1)]" %>
-                </li>
+                <%= content_tag(:li, link_to("Get Involved", home_page_path, class: class_names("block py-3 px-4 text-white"), "aria-current": "page"), class: class_names("block lg:inline hover:shadow-[0_3px_0_0_rgba(249,111,0,1)]", {active: current_page?(home_page_path)})) %>
```
littleforest, line 24: Should we just make this path as "#" for now until we have something real for this page? Or maybe we should just set `active: false` for now. I think it looks a little weird to have both links highlighted when on the home page.

littleforest (commented):
Nice! Just a question/idea about the Get Involved page.

app/views/layouts/_nav.html.erb
fatimaburke: I agree and have set active to false for now.

littleforest (approved):
👍

PR #201, opened 2023-05-17
Dynamically change count of selected Program or National Standard Type

littleforest:
Asana ticket: https://app.asana.com/0/1203289004376659/1204601802991345/f

This dynamically changes the count of the number of Program Type checkboxes checked and the National types checked. The Clear All is using a native Custom Event instead of a Stimulus method. I've tried unsuccessfully to use the Stimulus [dispatch](https://stimulus.hotwired.dev/reference/controllers#cross-controller-coordination-with-events) method or the Stimulus [outlet feature](https://stimulus.hotwired.dev/reference/outlets) for cross-controller communication. Would like to get this out now and can revisit the Custom Event code later.

[Reference used for implementing the Custom Event](https://fullstackheroes.com/tutorials/stimulus/create-custom-events/)

https://github.com/ApprenticeshipStandardsDotOrg/ApprenticeshipStandardsDotOrg/assets/1938665/e4b4b22c-f297-4462-bfa6-5b56c96b8ac4


app/assets/stylesheets/application.css
```diff
@@ -18,44 +18,42 @@
  @tailwind components;
  @tailwind utilities;
 
-
-
  .occupation-tabs *:focus, .occupation-tabs input:focus + label {
```
FerPerales, line 21: We may not need this change as part of this PR

Note: it seems we don't have an standard regarding indentation. We should set something to handle that now that there is five people in the project

app/javascript/controllers/checkbox_filter_controller.js
```diff
@@ -0,0 +1,24 @@
...
+  }
+
+  updateCount() {
+    const length = this.inputTargets.filter(checkbox => checkbox.checked).length
+    this.setCountValue(length)
+  }
+
+  setCountValue(length) {
+    this.countTarget.innerHTML = length
+    if (length > 0) {
```
FerPerales, line 18: We can use [toggle](https://developer.mozilla.org/en-US/docs/Web/API/Element/classList):

```
this.countWrapperTarget.classList.toggle(this.displayableClass, length > 0)
```

FerPerales:
@littleforest looks good. Left a couple of comments

fatimaburke:
All good on the design end.

fatimaburke (approved):

app/assets/stylesheets/application.css
littleforest: I can remove the indentation change.

app/javascript/controllers/checkbox_filter_controller.js
littleforest: Oh cool, I didn't know toggle could take an argument like that.

PR #210, opened 2023-05-18
Add contact form

FerPerales:
Asana ticket: https://app.asana.com/0/1203289004376659/1204452198889971/f

Form page:

image

Admin section to consult contact requests:
image

Admin show view for contact request:

image

### DIFF
--- a/app/policies/contact_request_policy.rb
+++ b/app/policies/contact_request_policy.rb
@@ -1,0 +1,5 @@
+class ContactRequestPolicy < AdminPolicy
+  def index?
+    admin_or_converter?
+  end
+end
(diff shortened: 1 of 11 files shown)


FerPerales:
@fatimaburke I couldn't find a design for this section. I took the freedom to search for a template, but we can update this with a design later

app/policies/contact_request_policy.rb
```diff
@@ -0,0 +1,5 @@
+class ContactRequestPolicy < AdminPolicy
+  def index?
+    admin_or_converter?
```
littleforest, line 3: Let's just have this be `admin?` The converters only need to deal with the Source Files and Data Imports.

littleforest (approved):
LGTM, just one minor Pundit change requested. Thanks for working on this!

app/policies/contact_request_policy.rb
FerPerales: @littleforest Thanks! I'll use the defaults from AdminPolicy, which exactly do that

PR #213, opened 2023-05-18
Show NOF badge only for Urban Institute

FerPerales:
Asana ticket: https://app.asana.com/0/1203289004376659/1204614576547722/f

Let's show the badge only for Urban Institute and NOF programs

image

### DIFF
--- a/app/models/occupation_standard.rb
+++ b/app/models/occupation_standard.rb
@@ -120,6 +120,12 @@ class OccupationStandard < ApplicationRecord
     ojt_type&.titleize
   end
 
+  def show_national_occupational_framework_badge?
+    national_occupational_framework? &&
+      organization_id.present? &&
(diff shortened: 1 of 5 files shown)


app/models/occupation_standard.rb
```diff
@@ -120,6 +120,12 @@ def ojt_type_display
     ojt_type&.titleize
   end
 
+  def show_national_occupational_framework_badge?
+    national_occupational_framework? &&
+      organization_id.present? &&
+      organization_id == Organization&.urban_institute&.id
```
littleforest, line 126:
```suggestion
      organization_id == Organization.urban_institute&.id
```
I don't think the first `&` is necessary?

littleforest (approved):
Thanks for picking this up!

PR #214, opened 2023-05-18
Show only Urban institute in occupational frameworks

FerPerales:
This ensures that only Occupation Standards linked to Urban Institute and of type `occupational_framework` are considered to be standards with National Occupational Framework.

### DIFF
--- a/app/models/organization.rb
+++ b/app/models/organization.rb
@@ -11,6 +11,6 @@ class Organization < ApplicationRecord
   end
 
   def self.urban_institute
-    @@urban_institute ||= Organization.find_by(title: "Urban Institute")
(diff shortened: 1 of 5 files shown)


app/models/organization.rb
```diff
@@ -11,6 +11,6 @@ def self.related_instructions_organizations(occupation_standard)
   end
 
   def self.urban_institute
-    @@urban_institute ||= Organization.find_by(title: "Urban Institute")
+    Organization.find_by(title: "Urban Institute")
```
littleforest, line 14: @FerPerales the memoization kept making the specs fail. I couldn't figure it out, so took it out for now. We can revisit next week.

PR #219, opened 2023-05-22
Setup dependabot by including a dependabot.yml file

sarahraqueld:
There are a number of vulnerabilities with the current versions of our dependencies. Setting up dependabot will help us keep up to date with security patches and other dependency updates.  

Add a dependabot.yml configuration file. Up to 3 PRs will be opened by dependabot at once.

### DIFF
--- a/.github/workflows/dependabot.yml
+++ b/.github/workflows/dependabot.yml
@@ -2,8 +2,4 @@ version: 2
 updates:
   - package-ecosystem: 'bundler'
     directory: '/'
-    schedule:
-      interval: 'weekly'
-      time: '08:00'
-      timezone: America/New_York
     open-pull-requests-limit: 3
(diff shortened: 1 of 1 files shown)


.github/workflows/dependabot.yml
```diff
@@ -0,0 +1,9 @@
+version: 2
+updates:
+  - package-ecosystem: 'bundler'
+    directory: '/'
+    schedule:
+      interval: 'weekly'
+      time: '08:00'
+      timezone: America/New_York
+    open-pull-requests-limit: 3
```
littleforest, line 9: Any reason not to have them just get created in real-time as they come up?

littleforest (commented):
Thanks for setting this up!

.github/workflows/dependabot.yml
sarahraqueld: Good question, I don't have a good one, will remove the schedule (will keep the open prs limit so we don't get flooded with more PRs than we can merge

sarahraqueld: done!

littleforest (approved):

PR #223, opened 2023-05-22
Update standards import mailer to go to Patrick

littleforest:
Asana ticket: https://app.asana.com/0/1203289004376659/1204651892096307/f

This removes the code that would notify an admin
on every new standards import (created from
scrapers, guest page, or internal admin upload).
    
The converters are currently actively working
on the site and do not need to be notified when
new files are added through the scrapers. In the
future we can add a separate mailer that gets
triggered when the scrapers add files.


sarahraqueld (approved):

sarahraqueld (approved):
I don't have a lot of context yet, but looks good to me

PR #227, opened 2023-05-23
Move dependabot config file to .github instead of .github/workflows

sarahraqueld:
This dependabot config file should live under .github instead of .github/workflows to work properly.


littleforest (approved):
Thanks!

FerPerales (approved):

PR #231, opened 2023-05-23
Add a schedule to dependabot config

sarahraqueld:
We're getting a failure on dependabot github actions:

```
The property '#/updates/0' did not contain a required property of 'schedule'
```

I had previously removed it in an attempt that not adding a schedule would allow dependabot open PRs continuously, but a schedule property is required.


sarahraqueld:
I'm going ahead and merging this as it is a small fix that does not affect application code, so Jeanine can go ahead and rebase her PR.

PR #235, opened 2023-05-23
Do not index occupation standards after commit until versions conflict is fixed

sarahraqueld:
We're getting some notifications on Rollbar when trying to index documents: 

`The client noticed that the server is not a supported distribution of Elasticsearch.`

I'm not sure how to fix it yet, so I'm removing this after commit callback so we stop getting these erros until this problem with versions if fixed.


sarahraqueld:
I believe I have another idea to try, so I'll close this PR in the meantime.

PR #244, opened 2023-05-25
Search Occupation Standards

sarahraqueld:
On previous PRs, we had added a basic elasticsearch configuration. Expand it so we can, in the future, perform a better search in multiple fields at the same time.


app/models/concerns/occupation_standard_search.rb
```diff
@@ -0,0 +1,33 @@
+module OccupationStandardSearch
```
sarahraqueld, line 1: I'll work on improving this module/refactoring it on next PRs

PR #252, opened 2023-05-29
Allow an admin to set a standard as public

FerPerales:
Asana ticket: https://app.asana.com/0/1203289004376659/1204653712611866/f

Add a checkbox only visible to admins to mark a standard as public from the Upload Apprenticeship Standards view

image

Also, it adds the same field in occupation standards show and edit view

image

### DIFF
--- a/app/controllers/standards_imports_controller.rb
+++ b/app/controllers/standards_imports_controller.rb
@@ -4,7 +4,7 @@ class StandardsImportsController < ApplicationController
   end
 
   def create
-    @standards_import = StandardsImport.new(create_params)
+    @standards_import = StandardsImport.new(standards_import_params)
 
     if @standards_import.save
       if user_signed_in?
@@ -24,7 +24,26 @@ class StandardsImportsController < ApplicationController
 
   private
 
-  def create_params
+  def standards_import_params
+    if current_user&.admin?
+      admin_create_params
+    else
+      user_create_params
+    end
+  end
+
+  def user_create_params
     params.require(:standards_import).permit(:name, :email, :organization, :notes, files: [])
+  end
+
+  def admin_create_params
+    params.require(:standards_import).permit(
+      :name,
+      :email,
+      :organization,
+      :notes,
+      :public_document,
+      files: []
+    )
   end
 end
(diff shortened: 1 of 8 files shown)


spec/models/occupation_standard_spec.rb
```diff
@@ -456,4 +456,19 @@
...
   end
+
+  describe "#public_document?" do
+    it "returns true if OccupationStandard#public_document flag is true" do
+      occupation_standard = build(:occupation_standard, public_document: true)
+
+      expect(occupation_standard.public_document?).to be true
+    end
+
+    it "trus true if associated standard import is public document regardless of public_document flag" do
```
sarahraqueld, line 467:
```suggestion
    it "returns true if associated standard import is public document regardless of public_document flag" do
```

spec/models/occupation_standard_spec.rb
```diff
@@ -456,4 +456,19 @@
...
+      occupation_standard = build(:occupation_standard, public_document: false)
+      allow_any_instance_of(OccupationStandard).to receive(:public_document?).and_return(true)
+
+      expect(occupation_standard.public_document?).to be true
```
sarahraqueld, line 471: It seems we are asserting on the same method we're stubbing above, which means, if we break the implementation, this test will not fail. Did you mean to stub the return of standards_import?


```suggestion
      allow_any_instance_of(StandardImport).to receive(:public_document?).and_return(true)

      expect(occupation_standard.public_document?).to be true
```

sarahraqueld (approved):
Looks great! Left a comment

spec/models/occupation_standard_spec.rb
FerPerales: Good catch! You are correct. I meant to stub StandardImport

app/controllers/standards_imports_controller.rb
```diff
@@ -24,7 +24,26 @@ def show
...
+  def admin_create_params
+    params.require(:standards_import).permit(
+      :name,
+      :email,
+      :organization,
+      :notes,
+      :public_document,
+      files: []
+    )
+  end
```
littleforest, line 48: I'm late to the game here, but a change for the future could be that we set up a Pundit policy for StandardsImport and use the policy to define the permitted parameters. See [SourceFile policy](https://github.com/ApprenticeshipStandardsDotOrg/ApprenticeshipStandardsDotOrg/blob/main/app/policies/source_file_policy.rb#L24-L29) and [controller](https://github.com/ApprenticeshipStandardsDotOrg/ApprenticeshipStandardsDotOrg/blob/main/app/controllers/admin/source_files_controller.rb#L21-L23) as an example.

PR #254, opened 2023-05-29
Use flipper flags to return similar programs conditionally

sarahraqueld:

In the sprit of keeping PRs short and easy to review, in this one, we're only introducing the usage of a flipper flag to conditionally return the results of the deprecated similar programs, or to call the new SimilarOccupationStandards class. 

In the next PR, I'll work on expanding/implementing the `SimilarOccupationStandards` class.

### DIFF
--- a/app/models/occupation_standard.rb
+++ b/app/models/occupation_standard.rb
@@ -166,11 +166,11 @@ class OccupationStandard < ApplicationRecord
   end
 
   def similar_programs
-    OccupationStandard.where(
-      "title ILIKE ?", "%#{self.class.sanitize_sql_like(title).split.join("%")}%"
-    ).where.not(
-      id: id
-    ).limit(MAX_SIMILAR_PROGRAMS_TO_DISPLAY)
+    if Flipper.enabled?(:similar_programs_elasticsearch)
+      SimilarOccupationStandards.similar_to(self)
+    else
+      similar_programs_deprecated
+    end
   end
 
   def ojt_type_display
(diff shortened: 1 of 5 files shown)


spec/support/helpers/stub_feature_flags.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+module Helpers
+  module StubFeatureFlags
+    # Enable/disable flipper for a specific feature for any actors
+    # stub_feature_flag(:staff_directory, true)
+    def stub_feature_flag(feature_name, value)
```
sarahraqueld, line 7: Adding only this method would not warranty the introduction of a whole Module, but from my experience, the project will need other helper methods, so including it now sets a good place for it to evolve

app/models/occupation_standard.rb
```diff
@@ -191,4 +191,12 @@ def show_national_occupational_framework_badge?
   def national?
     national_standard_type.present?
   end
+
+  def similar_programs_deprecated
```
sarahraqueld, line 195: This method will need to be cleaned up after the feature is validated and the feature flag is turned on

app/models/occupation_standard.rb
```diff
@@ -169,11 +169,11 @@ def related_instructions_hours_in_human_format
   end
 
   def similar_programs
-    OccupationStandard.where(
-      "title ILIKE ?", "%#{self.class.sanitize_sql_like(title).split.join("%")}%"
-    ).where.not(
-      id: id
-    ).limit(MAX_SIMILAR_PROGRAMS_TO_DISPLAY)
+    if Flipper.enabled?(:similar_programs_elasticsearch)
```
sarahraqueld, line 172: This will always return false until we log into the console and set it to true.

app/models/similar_occupation_standards.rb
```diff
@@ -0,0 +1,13 @@
...
+  def self.similar_to(occupation_standard)
+    new(occupation_standard).top_five
+  end
+
+  def initialize(occupation_standard)
+    @occupation_standard = occupation_standard
+  end
+
+  def top_five
+    OccupationStandard.search
```
littleforest, line 11: Does this need to limit to 5?

littleforest: Oh you are doing that in a future PR.

PR #255, opened 2023-05-30
Implement the "More Like this" query for occupation standards

sarahraqueld:

Refactor Previous work: We don't need the searchable and occupation standard search modules anymore. We'll only need the SimilarOccupationStandards class;
Include the elasticsearch-rails and elasticsearch-model gems;
Implement a "More Like This" ES query to return similar occupation standards.

"The More Like This Query finds documents that are "like" a given set of documents. In order to do so, MLT selects a set of representative terms of these input documents, forms a query using these terms, executes the query and returns the results. The user controls the input documents, how the terms should be selected and how the query is formed."

## Screenshots

With a Occupation Standard named "Pipe Fitter", we're able to get similar results when there's another standard named "Pipe Fitter level 2"
Screenshot 2023-05-30 at 11 13 05

Screenshot 2023-05-30 at 11 13 09

## Feature flag

This is behind a feature flag (`similar_programs_elasticsearch`), and will not get released once it's merged


spec/models/similar_occupation_standards_spec.rb
```diff
@@ -0,0 +1,18 @@
...
+      create(:occupation_standard, title: "Mechanic")
+      medical_assistant_standard = create(:occupation_standard, title: "Medical Assistant")
+      create(:occupation_standard, title: "Medical Assistant")
+      create(:occupation_standard, title: "Admin Assistant")
+
+      OccupationStandard.import
+      sleep 1
+      similars_to_medical_assistant = SimilarOccupationStandards.similar_to(medical_assistant_standard)
+
+      expect(similars_to_medical_assistant.pluck(:title)).to eq ["Admin Assistant", "Medical Assistant"]
```
sarahraqueld, line 15: This should actually be the other way around, but IDK why "Admin Assistant" is considered higher relevance by ES.

sarahraqueld: Fixed by adding a snowball analyzer to the query.

app/models/similar_occupation_standards.rb
```diff
@@ -8,6 +10,27 @@ def initialize(occupation_standard)
...
+    OccupationStandard.__elasticsearch__.search(query).records
+  end
+
+  private
+
+  def query
+    {
+      size: 5,
+      query: {
+        more_like_this: {
```
littleforest, line 22: Is this just a custom name you give the query?

app/models/similar_occupation_standards.rb
```diff
@@ -8,6 +10,27 @@ def initialize(occupation_standard)
...
   def top_five
-    OccupationStandard.search
+    OccupationStandard.__elasticsearch__.search(query).records
+  end
+
+  private
+
+  def query
+    {
+      size: 5,
```
littleforest, line 20: I'm wondering if we should make this a constant, and rename the `top_five` method name to be more generic in case we decide to change the number of results we want to return. Then you could also stub the constant in the spec to be something small and confirm that it returns the correct number of results.

littleforest (commented):
This is so exciting!

app/models/similar_occupation_standards.rb
sarahraqueld: No, this is a Elasticsearch specialized query: https://www.elastic.co/guide/en/elasticsearch/reference/7.17/query-dsl-mlt-query.html#query-dsl-mlt-query

sarahraqueld:
A next step I'm not sure how I'll implement yet is indexing the associations, so that we can have similar occupation standards when they have similar work processes, for instance.

spec/models/occupation_standard_spec.rb
```diff
@@ -199,13 +199,15 @@
     it "returns occupation standards that match the given query" do
       _mechanic = create(:occupation_standard, title: "Mechanic")
       medical_assistant = create(:occupation_standard, title: "Medical Assistant")
-
+      medical_assistant_ii = create(:occupation_standard, title: "Medical Assistant II")
+      sleep 1
```
sarahraqueld, line 203: This unfortunately is still necessary, If I remove it, the tests fail :(

spec/models/similar_occupation_standards_spec.rb
```diff
@@ -0,0 +1,19 @@
...
+  describe ".similar_to", :elasticsearch do
+    it "returns records similar to the provided, sorted by similarity score" do
+      create(:occupation_standard, title: "Mechanic")
+      medical_assistant_standard = create(:occupation_standard, title: "Medical Assistant")
+      create(:occupation_standard, title: "Medical Assistant II")
+      create(:occupation_standard, title: "Admin Assistant")
+      create(:occupation_standard, title: "Medical Assistant")
+
+      OccupationStandard.import
+      sleep 1
```
sarahraqueld, line 13: This unfortunately is still necessary, If I remove it, the tests fail :(

app/models/similar_occupation_standards.rb
sarahraqueld: done ✅

spec/models/similar_occupation_standards_spec.rb
sarahraqueld: Found a way to remove it :D

spec/models/occupation_standard_spec.rb
sarahraqueld: Found a way to remove it!

app/models/similar_occupation_standards.rb
littleforest: Does this method ensure that it never returns the initial input record? I skimmed the docs, but didn't see it explicitly mentioned, but would assume so.

sarahraqueld: Yes it does, and I have a spec covering this requirement

littleforest (approved):
Looks great!

PR #263, opened 2023-05-31
Add correct navbar to pages

jenlu33:
Asana ticket: [https://app.asana.com/0/1203289004376659/1204607522639589/f](https://app.asana.com/0/1203289004376659/1204607522639589/f)

I realized I was using the wrong header, so I replaced all of them with the correct one so the search form doesn't show up on these pages. This way we don't need the conditional for the occupation_standard header


sarahraqueld:
Hey! it seems the test failing it's a flaky one, so I'm opening a PR removing it so it doesn't block your work.

sarahraqueld (approved):
I don't have a lot of context into this, but code-wise, looks good to me

littleforest (commented):
Oh that was way easier than I thought it would be.

PR #264, opened 2023-05-31
Allow an admin to set a standard as public

FerPerales:
Asana ticket: https://app.asana.com/0/1203289004376659/1204653712611866/f

This PR moves public_document attribute from OccupationStandard to SourceFile


littleforest (approved):
Looks great, thanks for fixing!

PR #265, opened 2023-05-31
Remove flaky specs

sarahraqueld:
These tests is failing intermittently on CI, and so I'm removing it until we have a better solution.


FerPerales:
Wondering if the test that is failing is also a flaky one 🤔

sarahraqueld:
> Wondering if the test that is failing is also a flaky one 🤔

Seems like it! Removed it as well

FerPerales (approved):

littleforest:
I think it was a different one that was flaky. The one that returns ["Medical Assistant", "Medical Assistant II", "Admin Assistant"].

littleforest:
Oh nevermind, the whole file got removed.

littleforest (approved):

PR #277, opened 2023-06-07
National standard default oa

jenlu33:
Asana ticket: [https://app.asana.com/0/1203289004376659/1204601802991343/f](https://app.asana.com/0/1203289004376659/1204601802991343/f)

I added a task to create the national registration agency and also updated the validations so that a state is not required.


lib/tasks/deployment/20230607142125_create_national_registration_agency.rake
```diff
@@ -0,0 +1,15 @@
+namespace :after_party do
+  desc "Deployment task: create_national_registration_agency"
+  task create_national_registration_agency: :environment do
+    puts "Running deploy task 'create_national_registration_agency'"
+
+    RegistrationAgency.find_or_create_by!(
+      agency_type: :oa
+    )
```
littleforest, line 8: This will actually just pull the first RegistrationAgency that has agency type `:oa`, but we really want a RegistrationAgency with `state_id: nil` and `agency_type: :oa`, right?

```
    RegistrationAgency.find_or_create_by!(
      state_id: nil,
      agency_type: :oa
    )
```

jenlu33: Oh, yes! I overlooked that, will add it in

littleforest (approved):
🚢

PR #284, opened 2023-06-12
Update document link every 15 minutes

FerPerales:
Asana ticket: https://app.asana.com/0/1203289004376659/1204583650961646/f

Polls OccupationStandard show view every minute to update the document link.

Document link expiration reduced to 15 minutes

Not sure how to do an automated test for this, though


littleforest (commented):
Nice! I'm wondering if we should have the refresh period be less frequent and maybe make the urls expire a little longer? Like what about refreshing every 15 minutes, and having URLs expire in an hour? Just thinking that as more people use the site, that is going to be a lot of requests to the server.

littleforest (approved):
:shipit:

PR #290, opened 2023-06-14
Add source url and add to scrapers

jenlu33:
Asana ticket: [https://app.asana.com/0/1203289004376659/1204375229874464/f](https://app.asana.com/0/1203289004376659/1204375229874464/f)

I added a separate source url to StandardsImports and added it to the scrapers.

### DIFF
--- a/app/jobs/scraper/oregon_job.rb
+++ b/app/jobs/scraper/oregon_job.rb
@@ -38,7 +38,8 @@ class Scraper::OregonJob < Scraper::WatirJob
             organization: organization
           ).first_or_initialize(
             notes: "From Scraper::OregonJob",
-            public_document: true
+            public_document: true,
+            source_url: apprenticeship_url
           )
 
           if standards_import.new_record?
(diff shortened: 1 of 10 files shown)


app/jobs/scraper/oregon_job.rb
```diff
@@ -38,7 +38,8 @@ def perform
             organization: organization
           ).first_or_initialize(
             notes: "From Scraper::OregonJob",
-            public_document: true
+            public_document: true,
+            source_url: apprenticeship_url
```
littleforest, line 42: I'm wondering if this should be `apprenticeship_url + "apprenticeship-opportunities.aspx"` ?

littleforest (commented):
Looks good! Just one question.

app/jobs/scraper/oregon_job.rb
jenlu33: That makes sense. I'll add that in

littleforest (approved):
🚢

PR #306, opened 2023-06-19
Enable rollbar in staging

sarahraqueld:
We're having a 500 error when trying to use the forgot my password feature in staging. To debug it, we'd like to see the error reported on rollbar. Having rollbar enabled in staging will also help us find out about issues sooner before deploying the app to prod.

### DIFF
--- a/config/initializers/rollbar.rb
+++ b/config/initializers/rollbar.rb
@@ -4,7 +4,7 @@ Rollbar.configure do |config|
 
   config.access_token = ENV["ROLLBAR_ACCESS_TOKEN"]
 
-  unless Rails.env.production? || Rails.env.staging?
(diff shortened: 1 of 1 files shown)


FerPerales (approved):

sarahraqueld:
> @sarahraqueld Looks good! Heads up: we are using RAILS_ENV production for staging as well. We may want to change it to staging, but that probably involves some extra work and can be tackled in another PR

huh, in that case we wouldn't need this PR for the purpose I'm targetting (investigating the issue in recovering a password). The error should be reported in rollbar, but it isn't. Maybe there's something else not enabled for  rollbar to work with staging.

sarahraqueld:
Fer mentioned on slack we got rollbar working with staging but through Heroku add-on, so I'm closing this

PR #310, opened 2023-06-19
Truncate title for work process

FerPerales:
Asana ticket: https://app.asana.com/0/1203289004376659/1204828166480417/f

image

### DIFF
--- a/app/dashboards/work_process_dashboard.rb
+++ b/app/dashboards/work_process_dashboard.rb
@@ -1,6 +1,8 @@
 require "administrate/base_dashboard"
 
 class WorkProcessDashboard < Administrate::BaseDashboard
+  MAX_TITLE_LENGTH = 50
+
   ATTRIBUTE_TYPES = {
     id: Field::String,
     competencies: Field::HasMany,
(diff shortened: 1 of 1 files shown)


app/dashboards/work_process_dashboard.rb
```diff
@@ -50,6 +52,6 @@ class WorkProcessDashboard < Administrate::BaseDashboard
   COLLECTION_FILTERS = {}.freeze
 
   def display_resource(work_process)
-    work_process.title
+    work_process.title.truncate(MAX_TITLE_LENGTH)
```
sarahraqueld, line 55: I see the `title` field is not NOT NULL on the db, which means some work processes will have a nil title. We'll have an undefined method error then, should we protect ourselves against it?

sarahraqueld: oops, ok, there's a presence validation on the model, so we're fine

sarahraqueld:
It'd be nice to have a screenshot of how it's looking after the change, so we as reviewers can quickly validate the fix works :)

FerPerales:
@sarahraqueld adding one 😄

PR #311, opened 2023-06-19
Home Page: Popular Industries section

sarahraqueld:
Asana ticket: https://app.asana.com/0/1203289004376659/1204799580445040/f

Since we're doing a number of changes to this home page, I think it'd be nice to release all of them at once. So I'm putting the changes behind a feature flag that will only get enabled once everything is approved. 

Include a new family of colors to the tailwind utilities file;
Extract the old "Featured" section to its own partial. This will make it easier for us to remove it when the time comes. 
Implement the new "Popular Industries" section following design specs.

Please notice the font-family is not changed yet. I think it makes sense to change it when the redesign is closer to finish. 

## Screenshots

With the `updated_home` flag disabled (no changes)

Screenshot 2023-06-19 at 15 45 54

With the flag enabled:

Screenshot 2023-06-19 at 16 16 00
Screenshot 2023-06-19 at 16 15 48
Screenshot 2023-06-19 at 15 21 10


sarahraqueld:
Pendencies before this can be released: 
Implement (or use it) the count of apprenticeships for each industry;
Fix the small bug in the "Building Trades" svg.

jenlu33 (approved):

FerPerales (approved):

PR #313, opened 2023-06-20
Move the Active States section to the top of the home page sections

sarahraqueld:
Asana ticket: https://app.asana.com/0/1203289004376659/1204799580445042/f

We should make this section the first thing we feature on the homepage. The states are probably more important than Industries or Occupations.

Screenshot 2023-06-20 at 10 17 58


sarahraqueld:
Since this is a trivial change, I think it's fine to merge without a review. Doing so will help me avoid conflicts with future PRs.

PR #324, opened 2023-06-26
Popular Occupations Section

sarahraqueld:
Asana ticket: https://app.asana.com/0/1203289004376659/1204799581648895/f

The homepage should feature Popular Occupations similar to original wireframes based on the onet codes

Introduce a PopularOnetCodes query, and then use these codes to search for Occupation Standards to show on the home page.

Screenshot 2023-06-27 at 11 07 38

Reminder this whole new home page is behind a feature flag (updated_home)


app/views/pages/home.html.erb
```diff
@@ -111,6 +111,34 @@
...
+    <div class="py-8 px-4 mx-auto max-w-screen-xl sm:py-16 lg:px-6">
+      <div class="mb-8 lg:mb-16 flex justify-between">
+        <h2 class="section-title mb-4 text-4xl tracking-tight font-bold text-gray-900">Popular Occupations</h2>
+          <%= link_to occupation_standards_path, class: "font-bold text-accent-50 sm:text-lg uppercase" do %>
+            See All
+            <svg aria-label="right arrow" class="inline mb-1" xmlns="http://www.w3.org/2000/svg" width="24" height="24" fill="none" stroke="currentColor" stroke-linecap="round" stroke-linejoin="round" stroke-width="2" class="feather feather-arrow-right"><path d="M5 12h14M12 5l7 7-7 7" /></svg>
+          <% end %>
+      </div>
+      <div class="text-center space-y-8 md:grid md:grid-cols-2 lg:grid-cols-3 xl:grid-cols-4 md:gap-8 xl:gap-8 md:space-y-0">
+        <% @popular_onet_codes.each do |onet_code| %>
```
FerPerales, line 124: PopularOnetCodesQuery.run is doing a query to return four codes, but then we do a query per code to show, which sounds like a n+1 problem. We could try the following:

* Making PopularOnetCodesQuery.run returns the records and iterate through them in the view
* If we need PopularOnetCodesQuery.run to return only the onet_code, we can do `OccupationStandard.where(onet_code: PopularOnetCodesQuery.run).each ` to grab all the records in one query

app/queries/popular_onet_codes_query.rb
```diff
@@ -0,0 +1,9 @@
+class PopularOnetCodesQuery
+  def self.run(limit: 4)
```
FerPerales, line 2: What are your thoughts on extracting the number 4 into a constant?

app/views/pages/home.html.erb
sarahraqueld: > Making PopularOnetCodesQuery.run returns the records and iterate through them in the view

I agree, and this was definitively my first option, however, building this query proved to be very complex. I couldn't write something legible, and since we're only showing 4 occupation standards on the view, this isn't really a problem, as we'll only hit the db 4 times. Anyway, if you know of a way to build this query quickly, feel free to take a stab at it

app/queries/popular_onet_codes_query.rb
sarahraqueld: thanks, I missed it. done

FerPerales (approved):

PR #327, opened 2023-06-27
Query the most popular industries to be featured on the home page

sarahraqueld:
Asana ticket: https://app.asana.com/0/1203289004376659/1204856912757781/f

On [The homepage should feature Popular Industries similar to the original wireframes](https://app.asana.com/0/1203289004376659/1204799580445040/f), we've implemented the layout of the Popular Industries section. But it is still hardcoding the count of apprenticeships per industry. 

Implement the query to show the actual number of apprenticeships per industry.

Pending: We're currently hardcoding the same icon for all industries. Waiting for an answer on how I can get an icon per industry. 

Screenshot 2023-06-27 at 13 02 39


app/models/industry.rb
```diff
@@ -7,4 +7,11 @@ class Industry < ApplicationRecord
   validates :prefix, presence: true, uniqueness: {scope: :version}
 
   scope :current, -> { where(version: CURRENT_VERSION) }
+
+  def self.popular(limit: 4)
```
FerPerales, line 11: We can extract limit here as a constant

FerPerales (approved):
Looks good! Left one comment

PR #330, opened 2023-06-30
Search occupation standards by industry name

jenlu33:
Asana ticket: [https://app.asana.com/0/1203289004376659/1204799581648899/f](https://app.asana.com/0/1203289004376659/1204799581648899/f)

I changed OccupationStandardQuery quite a bit to use raw SQL because I could not find a way to be able to search the industry name while still using `:q` as the search parameter without doing so. Let me know if there is a better way to do this!

Screenshot 2023-06-30 at 1 32 20 PM

### DIFF
--- a/app/queries/occupation_standard_query.rb
+++ b/app/queries/occupation_standard_query.rb
@@ -15,14 +15,14 @@ class OccupationStandardQuery
       .by_state_abbreviation(search_term_params[:state_abbreviation])
       .by_national_standard_type(search_term_params[:national_standard_type]&.keys)
       .by_ojt_type(search_term_params[:ojt_type]&.keys)
-      .and(
-        occupation_standards.by_title(search_term_params[:q])
-        .or(
-          occupation_standards.by_rapids_code(search_term_params[:q])
-        )
-        .or(
-          occupation_standards.by_onet_code(search_term_params[:q])
-        )
+      .where(
+        "title ILIKE :q OR rapids_code ILIKE :q OR onet_code ILIKE :q OR
+        occupation_standards.id IN (
+          SELECT occupation_standards.id FROM occupation_standards
+          JOIN industries ON occupation_standards.industry_id = industries.id
+          WHERE industries.name ILIKE :q
+        )",
+        q: "%#{search_term_params[:q]}%"
       )
   end
 end
--- a/app/queries/occupation_standard_query.rb
+++ b/app/queries/occupation_standard_query.rb
@@ -15,14 +15,17 @@ class OccupationStandardQuery
       .by_state_abbreviation(search_term_params[:state_abbreviation])
       .by_national_standard_type(search_term_params[:national_standard_type]&.keys)
       .by_ojt_type(search_term_params[:ojt_type]&.keys)
-      .where(
-        "title ILIKE :q OR rapids_code ILIKE :q OR onet_code ILIKE :q OR
-        occupation_standards.id IN (
-          SELECT occupation_standards.id FROM occupation_standards
-          JOIN industries ON occupation_standards.industry_id = industries.id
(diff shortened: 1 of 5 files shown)


app/queries/occupation_standard_query.rb
```diff
@@ -15,14 +15,14 @@ def self.run(occupation_standards, search_term_params)
...
       .by_ojt_type(search_term_params[:ojt_type]&.keys)
-      .and(
-        occupation_standards.by_title(search_term_params[:q])
-        .or(
-          occupation_standards.by_rapids_code(search_term_params[:q])
-        )
-        .or(
-          occupation_standards.by_onet_code(search_term_params[:q])
-        )
+      .where(
```
FerPerales, line 18: We can mode this into a scope and only call it from here, similar to code above. This will also allow us to test this piece of code on isolation

app/queries/occupation_standard_query.rb
```diff
@@ -15,14 +15,14 @@ def self.run(occupation_standards, search_term_params)
...
+        "title ILIKE :q OR rapids_code ILIKE :q OR onet_code ILIKE :q OR
+        occupation_standards.id IN (
+          SELECT occupation_standards.id FROM occupation_standards
+          JOIN industries ON occupation_standards.industry_id = industries.id
+          WHERE industries.name ILIKE :q
+        )",
+        q: "%#{search_term_params[:q]}%"
```
littleforest, line 25: @jenlu33 I think we could leave the existing code here as-is, and add another `or` segment for searching by industry name like:

```
.and(
  occupation_standards.by_title(search_term_params[:q])
  .or(
    occupation_standards.by_rapids_code(search_term_params[:q])
  )
  .or(
    occupation_standards.by_onet_code(search_term_params[:q])
  )
  .or(
    occupation_standards.by_industry_name(search_term_params[:q])
  )
)
```

And then the `occupation_standards.by_industry_name` method could join with industries and try to match on the name. What do you think about that? Happy to pair together.

jenlu33: @littleforest I tried doing this before, but I would get this error:
`Relation passed to #or must be structurally compatible. Incompatible values: [:joins]`
When I looked it up, it seems like the issue is that the other scopes do not need a joins while the scope for `by_industry_name` does and it's causing problems when used with `or`

littleforest: We might be able to fake it by doing an isolated `join` query, and then returning a query that just grabs the ids. Like:

```
  scope :by_industry_name, ->(name) do
    if name.present?
      ids = OccupationStandard.joins(:industry).where("industries.name ILIKE ?", "%#{sanitize_sql_like(name).split.join("%")}%").pluck(:id)
      where(id: ids)   # <-- this will be the final scope that gets returned to be used with the "or" statement
    end
  end
```

jenlu33: Thanks, I'll try this when I get the chance!

spec/queries/occupation_standard_query_spec.rb
```diff
@@ -140,4 +140,20 @@
...
   end
+
+  it "allows searching by industry name" do
+    industry1 = create(:industry, name: "Healthcare Support Occupations")
+    industry2 = create(:industry, name: "Repair Occupations")
+
+    occupation_standard = create(:occupation_standard, industry: industry1)
+    create(:occupation_standard, industry: industry2)
+
+    params = {q: "Healthcare Support Occupations"}
```
littleforest, line 151: Would you mind changing this query to be something like: "healthcare support" so it also tests the case-insensitivity as well as the partial matching all at once? Otherwise, looks great!

littleforest (approved):
Just one minor request, otherwise looks good to go!

PR #334, opened 2023-07-11
Make ONET and RAPIDs code clickable labels

FerPerales:
Asana ticket: https://app.asana.com/0/1203289004376659/1205002769046394/f


crystal-williams-brown (approved):
The ticket doesn't have an set of requirements or meaning of done. Not sure what users are expecting when clicking the buttons. If they'd want to see an explanation of the code, other results with the same code, or more details of the standard they clicked the button on.

I think this looks good if they goal is returning a query for the o net or rapid code.

littleforest (approved):
:shipit:

PR #335, opened 2023-07-11
Accessibility fixes

crystal-williams-brown:
Asana ticket: https://app.asana.com/0/1203289004376659/1204974668381977/f

I made some edits to fix accessibility issues on the following pages
https://www.apprenticeshipstandards.org/occupation_standards
https://www.apprenticeshipstandards.org/occupation_standards/national_standards
What I did:

- Fixed broken aria-labeledby attributes. They were missing a matching element that would contain the aria information so I changed them to use an aria-label instead.
- Changed the hierarchy of h tags so they would be in level order, providing navigational guidance to users

Note: There were contrast issues that I didn't fix because they are for a disabled element and a hidden label, respectively.

Here are pictures of what the pages look like now.
Screenshot 2023-07-11 at 11 32 21 AM
Screenshot 2023-07-11 at 11 42 41 AM

### DIFF
--- a/app/views/occupation_standards/_header.html.erb
+++ b/app/views/occupation_standards/_header.html.erb
@@ -1,4 +1,4 @@
-<section aria-labelledby="page-title" class="occ-hero bg-no-repeat bg-cover bg-center bg-gray-700">
+<section aria-label="search" class="occ-hero bg-no-repeat bg-cover bg-center bg-gray-700">
   <%= render partial: "layouts/nav" %>
   <div class="max-w-screen-xl px-4 pt-8 pb-16 mx-auto lg:py-16">
 
--- a/app/views/occupation_standards/index.html.erb
+++ b/app/views/occupation_standards/index.html.erb
@@ -41,9 +41,9 @@
                 <div>
                     <div class="inline">
                         <% if occupation_standard.ojt_type %>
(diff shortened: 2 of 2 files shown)


FerPerales:
@crystal-williams-brown thanks for working on this! I still see the issues reported by Ax Dev Tools

Here's /occupation_standards

image

And here's /occupation_standards/national_standards

image

crystal-williams-brown:
I thought I had removed that id. I'll remove it now

crystal-williams-brown:
@FerPerales this is not the id I thought it was, that one is gone. I don't know where this ID is

crystal-williams-brown:
oh, I think it's from this
```
  <div id="<%= occupation_standard.ojt_type %>-occupation-standard" role="tooltip" class="tooltip invisible absolute z-10 inline-block rounded-lg bg-primary-800 px-3 py-2 text-sm font-medium text-white opacity-0 shadow-sm">
               <%= t("activerecord.attributes.occupation_standard.definitions.#{occupation_standard.ojt_type}").html_safe %>
               <div class="tooltip-arrow" data-popper-arrow></div>
  </div>
  ```
I'm not sure what this id is needed for

FerPerales:
@crystal-williams-brown it may not be needed, but worth looking if there's some CSS or JS associated to that selector before removing

crystal-williams-brown:
I added the `<%= occupation_standard.id %>` to make the target for the
tooltip uniq.

PR #337, opened 2023-07-12
Add suggestions in occupation search form

FerPerales:
Asana ticket: https://app.asana.com/0/1203289004376659/1204799581648902/f

https://github.com/ApprenticeshipStandardsDotOrg/ApprenticeshipStandardsDotOrg/assets/1335038/50126019-1eb7-455b-9cf1-8fdd0c362eca


littleforest (commented):
@FerPerales active_model_serializers is pretty old and I don't think will be getting updated. What do you think about using https://github.com/procore/blueprinter instead?

app/javascript/controllers/typeahead_controller.js
```diff
@@ -0,0 +1,44 @@
...
+export default class extends Controller {
+  // Input where the typeahead will be hooked
+  static targets = ["input"]
+  // Details:
+  // src: the data source. Must return an array. URL must include a wildcard, which will be replaced
+  // with the value from the input
+  // wildcard: the string used as a wildcard
+  // identifier: the key from the returned objects used to display as suggestion in the typeahead dropdown
+  static values = {
+    src: { type: String, default: "/occupation_standards.json?q=QUERY"},
```
littleforest, line 13: Can we not put a default, and instead use the route name on the Rails view side? I like to avoid hardcoding any routes in JS since it is easier to end up with broken links if we change routing.

littleforest (commented):
Thanks for working on this, @fer! What do you think about leaving the search result in the search box?

FerPerales:
> @FerPerales active_model_serializers is pretty old and I don't think will be getting updated. What do you think about using https://github.com/procore/blueprinter instead?

Sure! I've never used that. Taking a look

FerPerales:
> Thanks for working on this, @fer! What do you think about leaving the search result in the search box?

@littleforest

FerPerales:
> Thanks for working on this, @fer! What do you think about leaving the search result in the search box?

@littleforest I'll give a try. I guess I can pass the search argument as `q` in Turbo.visit

app/javascript/controllers/typeahead_controller.js
FerPerales: @littleforest I'd like to put defaults values as examples in case someone wants to reuse the controller. We can keep the defaults with "dummy" values and add the real ones in the view. What are your thoughts on this?

littleforest: Sure that sounds good!

littleforest (commented):
I love it! I'm wondering if I should retract my suggestion for including the query in the search results, because now if you click the search icon, it won't return anything because of the onet and/or rapids code in the search term. Is it easy to just make the search results show the occupation standard title by itself?

littleforest:
>  Is it easy to just make the search results show the occupation standard title by itself

Or since someone could be searching on onet code (instead of title), is it possible to just display in the search box the exact characters they ended up typing?

littleforest (approved):
Looks great! 🚢

PR #338, opened 2023-07-12
Elasticsearch updates

littleforest:

This is somewhat of a WIP, but is working well enough that I think we could merge this and flip the Flipper flag to use the ES results, and continue iterating.

Changes in this PR:
* Set index name based on environment so we can debug more easily
* Use 1 shard in test, dev environments to ensure consistent scoring results
* Use the `snowball` analyzer for the `work_process_titles` so we don't match on "and" and "or" stopwords
* Do not include as much of the occupation standard record data to be included in ES. Just include the fields we are indexing and matching on for now.
* Add a `debug` flag for the `similar_to` method. This can be cleaned up later.
* Add specs for the `similar_to` method

We weren't entirely sure what the `more_like_this` function was doing, as the specs pass with and without that clause. More work will be done investigating how to use the built-in `more_like_this` feature.

### DIFF
--- a/app/models/similar_occupation_standards.rb
+++ b/app/models/similar_occupation_standards.rb
@@ -1,19 +1,24 @@
 class SimilarOccupationStandards
-  attr_reader :occupation_standard
+  attr_reader :occupation_standard, :debug
 
   RESULTS_SIZE = 5
   MINIMUM_SCORE = 0.2
 
-  def self.similar_to(occupation_standard)
-    new(occupation_standard).similar_to
+  def self.similar_to(occupation_standard, debug = false)
+    new(occupation_standard, debug).similar_to
   end
 
-  def initialize(occupation_standard)
+  def initialize(occupation_standard, debug = false)
     @occupation_standard = occupation_standard
+    @debug = debug
   end
 
   def similar_to
-    OccupationStandard.__elasticsearch__.search(query).records
+    search = OccupationStandard.__elasticsearch__.search(query)
+    if debug
+      puts search.search.definition[:body][:query].to_json
+    end
+    search.records.to_a
   end
 
   private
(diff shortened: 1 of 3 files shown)


jenlu33 (approved):

crystal-williams-brown (approved):
Looks good to me

app/models/similar_occupation_standards.rb
```diff
@@ -1,19 +1,29 @@
...
-  def initialize(occupation_standard)
+  def initialize(occupation_standard, debug = false)
     @occupation_standard = occupation_standard
+    @debug = debug
   end
 
   def similar_to
-    OccupationStandard.__elasticsearch__.search(query).records
+    response = OccupationStandard.__elasticsearch__.search(query)
+    if debug
```
FerPerales, line 18: @littleforest is there a performance concern to conditionally print debugging output? I don't see an issue with this method always logging with Rails.logger.info and having this data available all the time. What are your thoughts?

littleforest: I feel like `Rails.logger.info` is not that helpful to me in debugging tests, since it's easier to use `puts` and see the output directly that having it go into the log. And I'd prefer to not have `puts` always on since it makes the test output noisy.

We are using the `bonsai` gem for the ES connection, and it actually takes care of the Elasticsearch.client connection setup. Ultimately I'd like to do the client setup manually, because then we can pass the logging flag directly in the Elasticsearch.client set up and trigger it that way.

I don't intend to keep this `if debug` code around forever, but since I was still going to be doing some more work on this class, I thought I would leave it for the short term.

FerPerales: If it's something temporal, we can keep it. Thanks for the clarification!

FerPerales (approved):

PR #340, opened 2023-07-12
Add Google reCAPTCHA to contact form and public standards import form

littleforest:
Asana tickets:
https://app.asana.com/0/1203289004376659/1204865513985057/f
https://app.asana.com/0/1203289004376659/1204865513985061/f

This adds the invisible reCAPTCHA v3 to the contact form and the public standards import form.

[Client side integration docs](https://developers.google.com/recaptcha/docs/v3)
[Server side integration docs](https://developers.google.com/recaptcha/docs/verify)

The default setup in the Google docs breaks the HTML form validations, so I followed [this SO post](https://stackoverflow.com/a/65645319/1753903) to reinstate the validations.

### DIFF
--- a/app/views/shared/_google_recaptcha.html.erb
+++ b/app/views/shared/_google_recaptcha.html.erb
@@ -1,6 +1,6 @@
 <script src="https://www.google.com/recaptcha/api.js"></script>
 <script>
   function onSubmit(token) {
-    document.getElementById(<%= form_id %>).submit();
+    document.getElementById("<%= form_id %>").submit();
   }
 </script>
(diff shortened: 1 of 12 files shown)


app/views/shared/_google_recaptcha.html.erb
```diff
@@ -0,0 +1,11 @@
+<script src="https://www.google.com/recaptcha/api.js"></script>
+<script>
+  function onSubmit(token) {
```
FerPerales, line 3: @littleforest wondering if we should wrap this inside `content_for :javascript` to not render the JS in the middle of the HTML

FerPerales (approved):

crystal-williams-brown (approved):
I haven't setup re-captcha before. this looks interesting

app/views/shared/_google_recaptcha.html.erb
littleforest: Tried this, but broke the reCAPTCHA so just going to merge as-is. Thanks for the suggestion.

PR #342, opened 2023-07-13
Remove stop word from work process title since sometimes matching

littleforest:
On CI we are not getting consistent scoring back
and it is resulting in a flaky test.

I added highlighting while debugging, and it seems that the "and" stop word is occasionally matching for some reason. I just ended up taking out the "and" in the work process title as a workaround. Don't love it, but seems to be working fine on production.

Other changes:
* The snowball analyzer seems to have been removed (but still seemed to work, so unsure), but I replaced the snowball analyzer with the english analyzer anyway as that [seemed to be the appropriate change](https://stackoverflow.com/a/41867221/1753903).
* Added `minimum_must_match` equal to 1 for the items in the should clause

### DIFF
--- a/app/models/occupation_standard.rb
+++ b/app/models/occupation_standard.rb
@@ -65,9 +65,9 @@ class OccupationStandard < ApplicationRecord
 
   settings(es_settings) do
     mappings dynamic: false do
-      indexes :title, type: :text, analyzer: :snowball
+      indexes :title, type: :text, analyzer: :english
       indexes :ojt_type, type: :text
-      indexes :work_process_titles, type: :text, analyzer: :snowball
+      indexes :work_process_titles, type: :text, analyzer: :english
       indexes :onet_code, type: :text, analyzer: :autocomplete
       indexes :rapids_code, type: :text, analyzer: :autocomplete
       indexes :national_standard_type, type: :text, analyzer: :keyword
--- a/app/models/similar_occupation_standards.rb
+++ b/app/models/similar_occupation_standards.rb
@@ -20,7 +20,6 @@ class SimilarOccupationStandards
       puts response.search.definition[:body][:query].to_json
       puts "HITS: #{response.results.total}"
       response.results.each do |result|
-        pp result
         puts "#{result._id}: #{result._score}"
       end
     end
@@ -60,11 +59,6 @@ class SimilarOccupationStandards
               }
             }
           ]
-        }
-      },
-      "highlight": {
-        "fields": {
-          "*": {}
         }
       }
     }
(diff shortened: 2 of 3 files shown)


jenlu33 (approved):

littleforest:
Screen Shot 2023-07-13 at 12 55 27 PM

Clearly I did not fix it 😭

littleforest:
Seems to sometimes still be matching on the "and" stopword (but not consistently):

```
#<Elasticsearch::Model::Response::Result:0x00007fa795b0c0b8
 @result=
  {"_index"=>"occupation_standards_test",
   "_type"=>"_doc",
   "_id"=>"01f806e8-3c64-400a-933b-69b5095482f4",
   "_score"=>32.91713,
   "_source"=>
    {"title"=>"Human Resource Specialist",
     "ojt_type"=>"hybrid",
     "onet_code"=>nil,
     "rapids_code"=>nil,
     "national_standard_type"=>nil,
     "state"=>"AL",
     "work_process_titles"=>
      ["Tracking and Management Systems", "Vehicle Inspection"]},
   "highlight"=>
    {"work_process_titles"=>["Tracking <em>and</em> Management Systems"]}}>
```

crystal-williams-brown:
@littleforest do you want to exclude "and"? If so, is there a [stop filter](https://www.elastic.co/guide/en/elasticsearch/reference/current/analysis-stop-tokenfilter.html)?

littleforest:
> do you want to exclude "and"
 
Yes, I was using the snowball filter, which should be doing that, but I just read that [it was removed](https://github.com/elastic/elasticsearch/issues/8690) awhile back (though I am confused how it (mostly) seems to be still working in Elasticsearch). I am trying to replace it with the `english` analyzer to see if that helps.

crystal-williams-brown:
> > do you want to exclude "and"
> 
> Yes, I was using the snowball analyzer, which should be doing that, but I just read that [it was removed](https://github.com/elastic/elasticsearch/issues/8690) awhile back (though I am confused how it (mostly) seems to be still working in Elasticsearch). I am trying to replace it with the `english` analyzer to see if that helps.

I think the analyzer should be a stop filter to exclude the stop words.

crystal-williams-brown:
https://medium.com/@andre.luiz1987/token-filters-analyzers-elasticsearch-eb7945fa7072
We have an analyzer in "app/models/occupation_standard.rb". It's in es_settings. I'm not sure if we have to setup something similar for "app/models/similar_occupation_standards.rb"

crystal-williams-brown (approved):
Looks good. Wish ES was easier to test (reliably).

PR #343, opened 2023-07-13
Reduce the boost on work_process_titles when getting similar results

littleforest:

We care mostly about title when finding similar programs.

### DIFF
--- a/app/models/similar_occupation_standards.rb
+++ b/app/models/similar_occupation_standards.rb
@@ -40,8 +40,7 @@ class SimilarOccupationStandards
             }},
             {match: {
               work_process_titles: {
-                query: occupation_standard.work_processes.pluck(:title).to_sentence,
-                boost: 5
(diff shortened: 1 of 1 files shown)


crystal-williams-brown (commented):
Hope I'm understanding correctly. You removed the boot for the work_process_titles so that the title will get preference over it. Could this be achieved by lowering the boost rather than removing it? Or does the work_process_title not really matter for matching?

littleforest:
> Could this be achieved by lowering the boost rather than removing it? Or does the work_process_title not really matter for matching?

@crystal-williams-brown great question. Removing the boost value essentially gives it a boost value of "1". We mostly care about matching on the title, so we are giving title a boost of 5. We do want to give records that match on work process title a higher score than records that don't match at all, so that's why we are reducing the boost value to 1.

crystal-williams-brown (approved):
Thank you for the explanation ^_^

PR #348, opened 2023-07-17
Build method to verify occupation standard hours match

crystal-williams-brown:
- add hours_meet_occupation_requirements? method
- Compare work_process_hours to occupation.time_based_hours
- If work_process_hours are greater than or equal to time_based_hours it returns true

Asana ticket: https://app.asana.com/0/1203289004376659/1204838413510054/f

Here is what it looks like
Screenshot 2023-07-18 at 4 22 55 PM

https://github.com/ApprenticeshipStandardsDotOrg/ApprenticeshipStandardsDotOrg/assets/99890957/afb0507d-4728-47a2-bc9b-64e3be21821f


spec/models/occupation_standard_spec.rb
```diff
@@ -510,4 +510,30 @@
       expect(occupation_standard.display_for_typeahead).to eq "Mechanic (15-1342) (9201)"
     end
   end
+
+  describe "#hours_meet_occupation_requirements?" do
+    it "returns true if work_process hours match occupation hours" do
+      occupation = create(:occupation, time_based_hours: "2000")
```
littleforest, line 516: Since your method is comparing the `work_process` hours only, let's make `time_based_hours` a different value (perhaps less than 2000), so the test is clear on what it is comparing against.

littleforest: Also, the hours should technically be an integer. Rails will adjust it, but it will avoid confusion if you put it in as the expected data type: `time_based_hours: 2000`

spec/models/occupation_standard_spec.rb
```diff
@@ -510,4 +510,30 @@
...
+      occupation_standard = create(:occupation_standard, occupation: occupation)
+      create(:work_process, maximum_hours: 2000, occupation_standard: occupation_standard)
+
+      expect(occupation_standard.hours_meet_occupation_requirements?).to be true
+    end
+
+    it "returns false if work_process hours do not match occupation hours" do
+      occupation = create(:occupation, time_based_hours: "2000")
```
littleforest, line 524:
```suggestion
      occupation = create(:occupation, time_based_hours: 2000)
```

littleforest (commented):
This looks good to me so far! Just some minor suggestions.

spec/models/occupation_standard_spec.rb
crystal-williams-brown: I'll fix the time_based hours. ojt hours were a string so I accidently carried that over to occupation through copy and past.

littleforest (commented):
@crystal-williams-brown this looks great! Can you also display the alert on the occupation_standards#show page?

littleforest (approved):
Looks great, thanks for adding the system specs. 🚢

PR #358, opened 2023-07-19
Remove OJT hours for competency based occupation

jenlu33:
Asana ticket: [https://app.asana.com/0/1203289004376659/1204649515289367/f](https://app.asana.com/0/1203289004376659/1204649515289367/f)

It looks like competency based occupation standards should not have OJT hours. I made it so that these don't show up in the views

Screenshot 2023-07-19 at 2 49 59 PM

### DIFF
--- a/app/models/work_process.rb
+++ b/app/models/work_process.rb
@@ -6,6 +6,8 @@ class WorkProcess < ApplicationRecord
 
   validates :title, presence: true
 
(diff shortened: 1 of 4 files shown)


littleforest (commented):
Looks good! Can you add a small system spec to the show view?

littleforest (approved):
🚢

PR #361, opened 2023-07-20
Fix query when state is null

FerPerales:
Asana ticket: https://app.asana.com/0/1203289004376659/1205072156506468/f


app/models/similar_occupation_standards.rb
```diff
@@ -29,37 +32,56 @@ def similar_to
   private
 
   def query
-    {
-      size: RESULTS_SIZE,
-      min_score: MINIMUM_SCORE,
```
FerPerales, line 34: @littleforest I couldn't find a way to integrate this using the DSL. Any idea?

app/models/similar_occupation_standards.rb
```diff
@@ -29,37 +33,56 @@ def similar_to
...
-          ],
-          minimum_should_match: 1,
-          must_not: [
-            {
-              term: {
-                _id: occupation_standard.id
-              }
-            }
-          ]
+    definition = Search.new
```
FerPerales, line 36: @littleforest I noticed that there's a more DSL way to do this instead of the imperative I'm doing. Not sure which one you had in mind

app/models/similar_occupation_standards.rb
```diff
@@ -29,37 +33,56 @@ def similar_to
...
+        }
+      )
+    )
+
+    if occupation_standard.registration_agency&.state
+      should.append(
+        Queries::Match.new(
+          state: {
+            query: occupation_standard.registration_agency&.state&.abbreviation
+          }
```
crystal-williams-brown, line 68: could this be turned into a method for getting the state abbreviation? Then you could call that method and make this line easier to read.

crystal-williams-brown (approved):
The DSL looks handy

app/models/similar_occupation_standards.rb
FerPerales: @crystal-williams-brown good point! I'll create a method in the model

PR #368, opened 2023-07-25
Add page titles

jenlu33:
Asana ticket: [https://app.asana.com/0/1203289004376659/1205065780366507/f](https://app.asana.com/0/1203289004376659/1205065780366507/f)

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -1,7 +1,7 @@
 <!DOCTYPE html>
 <html lang="en">
   <head>
-    <title>ApprenticeshipStandardsDotOrg</title>
+    <title><%= @page_title ? @page_title + " - ApprenticeshipStandardsDotOrg" : "ApprenticeshipStandardsDotOrg" %></title>
     <meta name="viewport" content="width=device-width,initial-scale=1">
     <%= csrf_meta_tags %>
     <%= csp_meta_tag %>
--- a/spec/system/pages/home_spec.rb
+++ b/spec/system/pages/home_spec.rb
@@ -20,6 +20,26 @@ RSpec.describe "pages/home" do
     expect(page).to_not have_link "Pipe Fitter", href: occupation_standard_path(pipe_fitter)
   end
 
+  it "displays the correct page title when page_title is present" do
+    allow(State).to receive(:find_by).and_return(build_stubbed(:state))
+    mechanic = create(:occupation_standard, :with_work_processes, :with_data_import, title: "Mechanic")
+
+    visit home_page_path
+
+    fill_in "q", with: "Mechanic"
+
+    find("#search").click
+
+    expect(page).to have_title "Occupations - ApprenticeshipStandardsDotOrg"
+  end
+
+  it "displays the correct page title when page_title is not present" do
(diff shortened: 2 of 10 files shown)


FerPerales:
@jenlu33 Looks good! It would be nice to add a couple of tests to make sure things render correctly when @page_title is present and when is not

app/views/layouts/application.html.erb
```diff
@@ -1,7 +1,7 @@
 <!DOCTYPE html>
 <html lang="en">
   <head>
-    <title>ApprenticeshipStandardsDotOrg</title>
+    <title><%= @page_title ? @page_title + " - ApprenticeshipStandardsDotOrg" : "ApprenticeshipStandardsDotOrg" %></title>
```
FerPerales, line 4: @jenlu33 if the controller does not define `@page_title`, this check will probably fail since the variable we are trying to access is not defined (need to verify with a test). We can do:

```
<%= defined?(@page_title) ? @page_title + " - ApprenticeshipStandardsDotOrg" : "ApprenticeshipStandardsDotOrg" %>
```

littleforest: @FerPerales if `@page_title` is not defined (is `nil`), then won't it just skip to the `else` part of the tertiary expression?

jenlu33:
@FerPerales Do you have insight on how I can fix these errors? I don't know where in our code a Chromedriver version is set so I'm not sure how to change it

FerPerales:
@jenlu33 yes. We have seen this error across our clients. 

Can you try this:

```
* Remove webdrivers from the Gemfile
* Update selenium-webdriver to the latest, 4.10.0
* Update capybara to the latest, 3.39.2
* Change the log level of Selenium from the default :info to :error (in test_helper.rb, Add Selenium::WebDriver.logger.level = :error
```


louis
:flag-gr:  [7 days ago](https://thoughtbot.slack.com/archives/C02C73BAV/p1689867979719439?thread_ts=1689787661.683109&cid=C02C73BAV)
In test/test_helper.rb:
Selenium::WebDriver.logger.level = :error

jenlu33:
@FerPerales I don't see a `test_helper.rb` file. Is that something I need to create?

FerPerales:
@jenlu33 right. We are using rspec. It should be rspec_helper.rb

Another solution that @crystal-williams-brown implemented for another client is adding this to `spec/rails_helper.rb`


```
Webdrivers::Chromedriver.required_version = "114.0.5735.90"
```

To use the previous version

spec/system/pages/home_spec.rb
```diff
@@ -20,6 +20,26 @@
     expect(page).to_not have_link "Pipe Fitter", href: occupation_standard_path(pipe_fitter)
   end
 
+  it "displays the correct page title when page_title is present" do
+    allow(State).to receive(:find_by).and_return(build_stubbed(:state))
+    create(:occupation_standard, :with_work_processes, :with_data_import, title: "Mechanic")
+
+    visit home_page_path
```
FerPerales, line 27: We can simplify this test if we go directly to `occupation_standards_path` and check that the title is correct. We can also remove the need to stub an State and create an occupation standard

FerPerales (approved):

PR #369, opened 2023-07-31
Allow admins to redact a PDF - Part 1

FerPerales:

image

image

First part of the feature:

* Integrate Foxit SDK
* Add Redact document in source file when a document is a PDF file


app/models/source_file.rb
```diff
@@ -34,4 +34,8 @@ def claimed?
   def standards_import
     active_storage_attachment.record
   end
+
+  def pdf?
+    active_storage_attachment&.filename&.extension == "pdf"
```
littleforest, line 39: What about calling content type on the attachment? `active_storage_attachment.blob.content_type == "application/pdf"`

config/environments/development.rb
```diff
@@ -64,6 +64,8 @@
   # Enable debug mode in development
   config.assets.debug = true
 
+  config.assets.compile = true
```
littleforest, line 67: What are the implications of this for making changes to assets in development? Do we have to restart the server every time? Should we make this a conditional for if we have "work on pdf redaction tickets flag turned on"?

config/routes.rb
```diff
@@ -1,4 +1,7 @@
 Rails.application.routes.draw do
+  namespace :admin do
+    get "redact_files/new"
+  end
```
littleforest, line 4: Do we need both these lines and line 41 below since it is not an AR-backed resource?

littleforest (commented):
This is great! Just a couple of small questions.

config/routes.rb
FerPerales: @littleforest good catch! We only need this in line 41

config/environments/development.rb
FerPerales: @littleforest We don't need this anymore. It was not the issue for missing assets

littleforest (approved):
🚢

PR #373, opened 2023-08-02
Remove webdrivers gem

littleforest:
The webdrivers gem is no longer necessary as Selenium Manager now handles its own installations of the Chrome Driver.

https://github.com/rails/rails/pull/48847


FerPerales (approved):
:dead-code-society:

jenlu33 (approved):

PR #379, opened 2023-08-11
Exclude Occupation Standards with No Work Processes

jenlu33:
Asana ticket: [https://app.asana.com/0/1203289004376659/1205072158398816/f](https://app.asana.com/0/1203289004376659/1205072158398816/f)

The popular occupations and industries on the home page were showing a count of all `occupation_standards` even though we filter out the ones without `work_processes` in search. 

I adjusted the queries to only include those with associated `work_processes`. I also added a `with_work_processes` scope to make sure we count the correct occupations

### DIFF
--- a/app/models/industry.rb
+++ b/app/models/industry.rb
@@ -10,9 +10,9 @@ class Industry < ApplicationRecord
   scope :current, -> { where(version: CURRENT_VERSION) }
 
   def self.popular(limit: POPULAR_LIMIT)
-    Industry.left_joins(:occupation_standards)
+    Industry.joins(occupation_standards: :work_processes)
       .group("industries.id")
-      .order("COUNT(occupation_standards.id) DESC")
+      .order(Arel.sql("COUNT(DISTINCT occupation_standards.id) DESC"))
       .limit(limit)
   end
 end
--- a/spec/queries/popular_onet_codes_query_spec.rb
+++ b/spec/queries/popular_onet_codes_query_spec.rb
@@ -10,7 +10,7 @@ RSpec.describe PopularOnetCodesQuery do
 
     stub_const("PopularOnetCodesQuery::LIMIT", 4)
 
-    result = described_class.run(limit: 3)
+    result = described_class.run
 
     expect(result).to eq ["33", "22", "11"]
   end
--- a/spec/queries/popular_onet_codes_query_spec.rb
+++ b/spec/queries/popular_onet_codes_query_spec.rb
@@ -2,9 +2,10 @@ require "rails_helper"
 
 RSpec.describe PopularOnetCodesQuery do
   it "returns the onet codes for popular occupation standards" do
-    create(:occupation_standard, title: "Medical Assistant", onet_code: "11")
-    create_list(:occupation_standard, 2, title: "Pipe Fitter", onet_code: "22")
-    create_list(:occupation_standard, 3, title: "Nurse", onet_code: "33")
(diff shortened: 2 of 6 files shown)


spec/queries/popular_onet_codes_query_spec.rb
```diff
@@ -2,9 +2,10 @@
 
 RSpec.describe PopularOnetCodesQuery do
   it "returns the onet codes for popular occupation standards" do
-    create(:occupation_standard, title: "Medical Assistant", onet_code: "11")
-    create_list(:occupation_standard, 2, title: "Pipe Fitter", onet_code: "22")
-    create_list(:occupation_standard, 3, title: "Nurse", onet_code: "33")
+    create(:occupation_standard, :with_work_processes, title: "Medical Assistant", onet_code: "11")
+    create_list(:occupation_standard, 2, title: "HR Specialist", onet_code: "12")
```
littleforest, line 6: One minor request, just in an effort to create as few DB records as possible in tests:

Could you just create one record instead of 2, which should be sufficient as we just want to test that onet_code "12" doesn't show up in the list of our top 4:

```suggestion
    create(:occupation_standard, title: "HR Specialist", onet_code: "12")
```

And then to ensure that our test is always testing that we would like at most 4 onet codes returned, can you add:
```
stub_const("PopularOnetCodesQuery:LIMIT", 4) 
```
so that if this constant is ever changed to less than 4, we still want our test to work as if it had a limit of 4. I.e. without stubbing the constant, if LIMIT is changed to 3, our test would still pass, but it would no longer be testing that the "12" is not getting returned due to not having work processes, it would end up not getting returned because we were just requesting a limit of 3. But if we fix the LIMIT to 4 in our test regardless, then we can be sure that we are testing the correct thing.

app/models/industry.rb
```diff
@@ -10,9 +10,9 @@ class Industry < ApplicationRecord
   scope :current, -> { where(version: CURRENT_VERSION) }
 
   def self.popular(limit: POPULAR_LIMIT)
-    Industry.left_joins(:occupation_standards)
+    Industry.joins(occupation_standards: :work_processes)
       .group("industries.id")
-      .order("COUNT(occupation_standards.id) DESC")
+      .order(Arel.sql("COUNT(DISTINCT occupation_standards.id) DESC"))
```
littleforest, line 15: nice!

spec/queries/popular_onet_codes_query_spec.rb
```diff
@@ -2,11 +2,14 @@
...
-    create_list(:occupation_standard, 2, title: "Pipe Fitter", onet_code: "22")
-    create_list(:occupation_standard, 3, title: "Nurse", onet_code: "33")
+    create(:occupation_standard, :with_work_processes, title: "Medical Assistant", onet_code: "11")
+    create(:occupation_standard, title: "HR Specialist", onet_code: "12")
+    create_list(:occupation_standard, 2, :with_work_processes, title: "Pipe Fitter", onet_code: "22")
+    create_list(:occupation_standard, 3, :with_work_processes, title: "Nurse", onet_code: "33")
     create(:occupation_standard, onet_code: nil)
 
+    stub_const("PopularOnetCodesQuery::LIMIT", 4)
+
```
littleforest, line 12: For line 13 below, can we run this without passing `limit: 3`, so it runs with the default limit of 4, therefore testing that the occupation standard with onet code 12 is not getting returned because it doesn't have a work process, not because it was 4th to be returned when our limit was only 3?

littleforest (approved):
🚢

PR #380, opened 2023-08-14
Admin can redact PDFs

FerPerales:

Admin can click Redact document when a Source File is a PDF document

image

This will take the user to the FoxIt SDK editor, where it will have the Source File opened and ready to be redacted:

image

Once the document is redacted, you can save the changes to all Occupation Standards that are related to the Source File by clicking the stamp button and then the "Save Redacted Document" option:

image

You will get an alert confirmation if the file was saved successfully of an error message if it didn't:

image

After confirmation, you will be redirected to previous page.

To confirm the changes worked, you can go to any Occupation Standard show view that is associated to the Source File you just redacted and check that we now have a Redacted file option:

image

Open the file and verify that this is the redacted version you uploaded:

image


spec/requests/admin/redact_file_spec.rb
```diff
@@ -51,4 +51,46 @@
...
+            expect(response).to be_successful
+          end
+        end
+
+        context "with redacted file" do
+          it "returns http success" do
+            admin = create(:admin)
+            active_storage_attachment = create(:active_storage_attachment)
+            source_file = create(:source_file, active_storage_attachment: active_storage_attachment)
+            redacted_file = fixture_file_upload("spec/fixtures/files/pixel1x1.jpg", "image/jpeg")
```
littleforest, line 80: I don't think you need the whole path, and just `pixel1x1.jpg` is sufficient.

littleforest (approved):
Can you add a little tutorial to Notion on how to do the redaction?

PR #383, opened 2023-08-17
Move search to Elasticsearch

littleforest:
This sets up the option to run all of the searching and filtering through Elasticsearch. This is initially set up to be behind a feature flag, but after it is established to be working correctly, then the old search through the Postgres database should be removed.

The analyzer for the occupation standard title and industry name was modified to use ngrams so that we could more easily do partial matching. The new analyzer uses the English stop words and 3-character n-grams.

This also moves several system specs over to view specs instead.


FerPerales (approved):
Thanks for working on this. Looks good! 🚢

PR #384, opened 2023-08-17
Fix Foxit Editor not loading PDFs from S3

FerPerales:

Use `rails_storage_proxy_url(object)` helper instead of `object.url`. This [helper](https://edgeguides.rubyonrails.org/active_storage_overview.html#proxy-mode) make Rails grab the content from S3 and serve it instead of redirecting to the S3 file, which is causing an issue.

I tested this doing the following:

* Get into staging rails console
* Grab one SourceFile record
* Call `app.rails_blob_url(@source_file.active_storage_attachment)`
* Copy the URL
* In FoxitSDK editor, click on `Open Local File -> Open URL File` and paste the URL

PDF can be opened now 🎉


littleforest (approved):
Nice find!

PR #387, opened 2023-08-23
Set up Docker deploy

littleforest:
The combination of the assets (in particular the Foxit SDK) and the Google chrome buildpacks put the Heroku slug at over 500MB. This uses a custom Docker deployment instead so as to not be limited by the slug size.

Working Chrome!

Screen Shot 2023-08-22 at 5 45 30 PM


crystal-williams-brown (approved):
Looks good

FerPerales (approved):

PR #390, opened 2023-08-23
Clickable states

jenlu33:
Asana ticket: [https://app.asana.com/0/1203289004376659/1205143399865539/f](https://app.asana.com/0/1203289004376659/1205143399865539/f)
 
I created a new `popular` method for states so that they could be sorted the same way the other popular sections are being sorted on the home page. Then I also made the states clickable so that it shows all the `occupation_standards` in that state.


littleforest (approved):
LGTM!

app/views/pages/home.html.erb
```diff
@@ -28,30 +28,22 @@
...
-          <h3 class="text-xl font-bold text-primary-700 mt-4">Massachusetts DAS</h3>
-        </div>
-        <div class="p-6 rounded-lg flex flex-col justify-center items-center">
-          <div>
-            <%= image_tag("Wisconsin-DWD.png", alt: "Wisconsin DWD") %>
-          </div>
-          <h3 class="text-xl font-bold text-primary-700 mt-4">Wisconsin DWD</h3>
-        </div>
+        <% @popular_states.each do |state| %>
+          <%= link_to occupation_standards_path(state: state.abbreviation) do %>
```
littleforest, line 32: @jenlu33 would you mind making a follow up PR that passes the parameters as: `q: "state:#{state.abbreviation}"`? We are accepting state as a parameter on its own, but if it is passed within `params[:q]`, then it will show up in the search box:

Screen Shot 2023-08-23 at 12 59 46 PM

PR #392, opened 2023-08-24
Change state parameter to :q

jenlu33:
Changes how the state parameter is being passed so that it shows up in the search bar
image


littleforest (approved):
👍

PR #401, opened 2023-08-30
Add service to query Onet Web Services API for related job titles

littleforest:
Asana ticket: https://app.asana.com/0/1203289004376659/1205360607082020/f

This adds a new service that queries the [Onet Web Services API](https://services.onetcenter.org/reference/online/occupation#overview) for related job titles. When the ONET scraper is run, it will also query this API for the related job titles. We also added an After Party task so existing ONET code records can be updated.

### DIFF
--- a/app/services/onet_web_service.rb
+++ b/app/services/onet_web_service.rb
@@ -1,0 +1,22 @@
+class OnetWebService
+  attr_reader :onet, :base_url
+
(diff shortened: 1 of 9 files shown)


crystal-williams-brown (approved):
looks good to me

jenlu33 (approved):

PR #408, opened 2023-09-07
Add a search analyzer to be used with the autocomplete searches

littleforest:
Asana ticket: https://app.asana.com/0/1203289004376659/1205398471115381/f

The autocomplete analyzer at index time breaks up the term into edge ngrams from 3-20. We had been also using the autocomplete analyzer at search time, but this was causing problems as a search term of "tree" would return "tree trimmer" but would also return "truck driver". Per the [Elasticsearch docs](https://www.elastic.co/guide/en/elasticsearch/reference/7.17/analysis-edgengram-tokenizer.html#_example_configuration_7), we should be using a separate analyzer to use at search time:

> Usually we recommend using the same analyzer at index time and at search time. In the case of the edge_ngram tokenizer, the advice is different. It only makes sense to use the edge_ngram tokenizer at index time, to ensure that partial words are available for matching in the index. At search time, just search for the terms the user has typed in, for instance: Quick Fo

We created a new search analyzer to be used with the autocomplete indexing analyzer that uses the standard tokenizer with a lowercase filter that does not overmatch results.

Example  before change:

Screen Shot 2023-09-07 at 11 57 17 AM

Example after change:

Screen Shot 2023-09-07 at 11 58 15 AM
Screen Shot 2023-09-07 at 11 58 20 AM

### DIFF
--- a/app/models/occupation_standard.rb
+++ b/app/models/occupation_standard.rb
@@ -63,6 +63,7 @@ class OccupationStandard < ApplicationRecord
           },
           autocomplete_search: {
             tokenizer: "standard",
+            filter: ["lowercase"],
             char_filter: ["my_char_filter"]
           }
         }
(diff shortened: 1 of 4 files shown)


jenlu33 (approved):
🎉

FerPerales (approved):

PR #414, opened 2023-09-12
Fix occupational frameworks count

FerPerales:
Asana ticket: 

This allows  `OccupationStandard.by_national_standard_type` to receive a single value as a symbol and work as expected. Previously it only worked with the value as string, since it is the format we search from the query form checkbox

This is how is called from home.html.erb, which display an incorrect result:

```
 <p class="font-bold text-accent-100">
   <%= OccupationStandard.by_national_standard_type(:occupational_framework).count %> Apprenticeships
 </p>
```


littleforest:
Eventually we should move these to get all records from ES so we have consistent data.

FerPerales:
@littleforest I would keep them in PG

PR #417, opened 2023-09-18
added back chromdriver download

crystal-williams-brown:
Asana ticket: https://app.asana.com/0/1203289004376659/1205523361121486/f


littleforest:
Going to close this for now as it seems like the Chrome error in the logs was a blip. Job seems to have run successfully since then.

PR #419, opened 2023-09-18
Add synonym filter to English analyzer

littleforest:
Asana ticket: https://app.asana.com/0/1203289004376659/1205503615126500/f

This rebuilds the English analyzer to add a synonyms filter. If the list gets long it is recommended to use a file instead, but this file must live on the Elasticsearch server, not the Rails server.


FerPerales (approved):
Nice! 💯 

I'm learning a lot about ES

PR #427, opened 2023-09-20
Refine Elasticsearch results

littleforest:
Asana tickets: 
https://app.asana.com/0/1203289004376659/1205503615126496/f
https://app.asana.com/0/1203289004376659/1205444475106625/f

When a user enters a search term, they are often returned results unrelated to their intended search due to related words. For example, searching "software developer" will also return results for "Child Development Teacher" because of the word "developer" in the search. To refine the elasticsearch results, when a user enters a search term, we look up the ONET code of the top hit and do a second search that only returns results matching the ONET prefix. In that way, "Child Development Teacher" would be excluded from a search for "software developer" since they do not share the same ONET code prefix.

**Note**
The Elasticsearch gem does not have a filter for "match", even though it is valid Elasticsearch syntax, so a Filters Match model was added. Ordinarily a match would not likely be in the filter section, since matches can be fuzzy and scoring would be important. However in the case of the `onet_prefix` analyzer, we are using `match` in order to have the onet code in the search term get tokenized identically to the `onet_code.prefix` being stored. We could use the existing "term" filter if we performed our own regex on the onet_code search term, but I prefer to use the tokenizer on both the input and the search term for consistency, just in case we make any other changes to it down the line.

### DIFF
--- a/app/controllers/occupation_standards_controller.rb
+++ b/app/controllers/occupation_standards_controller.rb
@@ -4,13 +4,7 @@ class OccupationStandardsController < ApplicationController
 
     if Flipper.enabled?(:use_elasticsearch_for_search)
       if search_term_params[:q].present?
-        es_response_0 = OccupationStandardElasticsearchQuery.new(
-          search_params: search_term_params
-        ).call
-        first_hit = es_response_0.records.first
-        if first_hit && first_hit.onet_code.present?
-          search_term_params[:onet_prefix] = first_hit.onet_code
-        end
+        refine_search_params
       end
 
       es_response = OccupationStandardElasticsearchQuery.new(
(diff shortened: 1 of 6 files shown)


app/controllers/occupation_standards_controller.rb
```diff
@@ -3,6 +3,10 @@ def index
     @page_title = "Occupations"
 
     if Flipper.enabled?(:use_elasticsearch_for_search)
+      if search_term_params[:q].present?
+        refine_search_params
```
littleforest, line 7: @FerPerales the only thing I'm a little worried about with this is if someone enters "15" and hits return. There could be matches against ONET "15-xxxx" and also RAPIDS code "15XX", and it would take the first hit and then grab that ONET code to limit on, thereby excluding results from one or the other. Should we only refine the search if the first character is a letter?

FerPerales: @littleforest good point. Should we match only when string is "15-" and not only "15"?

littleforest: What about not worrying about the numbers, and only doing search refinement when someone enters letters for search? That way if someone enters "15", they would still get records back for ONET codes 15-xxxx and RAPIDS codes 15XX.

FerPerales (changes requested):
🚢

FerPerales (approved):

PR #432, opened 2023-09-26
Add related_job_titles to admin/occupation_standard show view

jenlu33:
Asana ticket: [https://app.asana.com/0/1203289004376659/1205444475106613/f](https://app.asana.com/0/1203289004376659/1205444475106613/f)

### DIFF
--- a/app/dashboards/occupation_standard_dashboard.rb
+++ b/app/dashboards/occupation_standard_dashboard.rb
@@ -18,7 +18,7 @@ class OccupationStandardDashboard < Administrate::BaseDashboard
     rapids_code: Field::String,
     registration_agency: Field::BelongsTo.with_options(scope: -> { RegistrationAgency.includes(:state) }),
     related_instructions: Field::HasMany,
-    related_job_titles: Field::String,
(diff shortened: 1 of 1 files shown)


FerPerales:
@jenlu33 this looks good, but there's a broken test.

Administrate expect all the fields to be database fields and, based on the error, it's trying to search on that new field, which is not really a db field.

If we don't care about search using this field, we can tell administrate to avoid searching on this field with this change:


```
related_job_titles: Field::String.with_options(searchable: false),
```

That should make the test pass.

jenlu33:
> @jenlu33 this looks good, but there's a broken test.
> 
> Administrate expect all the fields to be database fields and, based on the error, it's trying to search on that new field, which is not really a db field.
> 
> If we don't care about search using this field, we can tell administrate to avoid searching on this field with this change:
> 
> ```
> related_job_titles: Field::String.with_options(searchable: false),
> ```
> 
> That should make the test pass.

Great, thank you! I was trying to google way to fix the specs and I couldn't figure it out

littleforest (approved):
:shipit:

FerPerales (approved):

PR #433, opened 2023-09-27
Use occupation titles for typeahead search

littleforest:
Asana ticket: https://app.asana.com/0/1203289004376659/1205523364822698/f

This uses the titles for typeahead search, but still searches the occupation standards once a title is selected. We are using the occupation titles for searching so users can see the whole universe of possible titles, not just the ones we happen to have standards for.

### DIFF
--- a/spec/system/occupation_standards/index_spec.rb
+++ b/spec/system/occupation_standards/index_spec.rb
@@ -285,48 +285,6 @@ RSpec.describe "occupation_standards/index" do
       expect(page).to have_link "Pipe Fitter"
     end
 
-    it "shows suggestions based on occupation title", :js do
-      mechanic = create(:occupation_standard, :with_work_processes, :with_data_import, title: "Mechanic")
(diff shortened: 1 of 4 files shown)


spec/system/occupation_standards/index_spec.rb
```diff
@@ -285,48 +285,6 @@
...
-      visit occupation_standards_path
-
-      expect(page).to_not have_selector "div", class: "tt-suggestion"
-
-      fill_in "q", with: "9108"
-
-      expect(page).to have_selector "div", class: "tt-suggestion", text: mechanic.display_for_typeahead
-      expect(page).to_not have_selector "div", class: "tt-suggestion", text: pipe_fitter.display_for_typeahead
-    end
-
```
littleforest, line 329: This section is for when the Flipper flag `use_elasticsearch_for_search` is disabled, except we do not have this flag in use in the occupations controller, so these specs are being removed.

FerPerales (approved):

PR #436, opened 2023-09-27
Reduce headline size

FerPerales:
Asana ticket: https://app.asana.com/0/1203289004376659/1205542807244423/f

We now use Digest::SHA2.hexdigest to digest the headline value we use to do ES collapse


littleforest (approved):
🎉

PR #437, opened 2023-09-27
Remove code related to typeahead for occupation standards

littleforest:
Asana ticket: https://app.asana.com/0/1203289004376659/1205523364822698/f

The typeahead searching was moved to the Occupation model, so this removes the old code which was using the OccupationStandard model.

### DIFF
--- a/spec/queries/occupation_standard_elasticsearch_query_spec.rb
+++ b/spec/queries/occupation_standard_elasticsearch_query_spec.rb
@@ -313,12 +313,12 @@ RSpec.describe OccupationStandardElasticsearchQuery, :elasticsearch do
     params = {q: "usER expERience"}
     response = described_class.new(search_params: params).call
 
-    expect(response.records.pluck(:id)).to eq [os1.id, os2.id]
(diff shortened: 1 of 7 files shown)


spec/queries/occupation_standard_elasticsearch_query_spec.rb
```diff
@@ -313,12 +313,12 @@
     params = {q: "usER expERience"}
     response = described_class.new(search_params: params).call
 
-    expect(response.records.pluck(:id)).to eq [os1.id, os2.id]
+    expect(response.records.pluck(:id)).to match_array [os1.id, os2.id]
```
littleforest, line 316: These are now matching in the reverse order. One reason is that the typeahead part of the query has been removed, so it isn't giving any extra points. The other reason is that for some reason the "user experience" query is matching on "designer" in the "ux designer" and giving it bonus points. I can't figure that out (seems like an error in ES), so I just switched this to `match_array` for now.

FerPerales (approved):

PR #438, opened 2023-09-28
Add Occupation table

jenlu33:
Asana ticket: [https://app.asana.com/0/1203289004376659/1205551095728739/f](https://app.asana.com/0/1203289004376659/1205551095728739/f)

Adds Occupations to the Admin dashboard. I added the ability to edit, but I'm not sure if we need to do that?


littleforest (approved):
I think having Edit available is fine. Looks good!

PR #439, opened 2023-09-28
Use occupation onet code from typeahead

FerPerales:
Asana ticket: https://app.asana.com/0/1203289004376659/1205503615126506/f

This adds the onet_code from an occupation selected in the typeahead to be part of the search query.


littleforest (approved):
Nice! Do you understand why it didn't work the other way?

FerPerales:
@littleforest yes. The element we were trying to click was not the one that triggers the event.

BTW, I spent some time trying to remove the flakiness of the test and did not have any luck. It may be related to ES since I found other ES-related tests that failed locally when I ran this 100 times

PR #445, opened 2023-10-05
Add rake task to import 2018 Onet codes

littleforest:
Asana ticket: https://app.asana.com/0/1203289004376659/1205644361039893/f

### DIFF
--- a/lib/tasks/deployment/20231005170246_import_2018_onet_codes.rake
+++ b/lib/tasks/deployment/20231005170246_import_2018_onet_codes.rake
@@ -1,0 +1,20 @@
+namespace :after_party do
+  desc "Deployment task: import_2018_onet_codes"
+  task import_2018_onet_codes: :environment do
+    puts "Running deploy task 'import_2018_onet_codes'"
+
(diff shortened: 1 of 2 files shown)


lib/tasks/deployment/20231005170246_import_2018_onet_codes.rake
```diff
@@ -0,0 +1,20 @@
+namespace :after_party do
+  desc "Deployment task: import_2018_onet_codes"
+  task import_2018_onet_codes: :environment do
```
crystal-williams-brown, line 3: How long does this task take? Is there any risk of timeout?

crystal-williams-brown (approved):
I'm curious about the risk of timeout or how to know if the task successfully completed. Otherwise I think this looks fine

lib/tasks/deployment/20231005170246_import_2018_onet_codes.rake
littleforest: Good question -- it's actually really fast, there shouldn't be any timeout issues.

PR #450, opened 2023-10-09
Add support for dynamic synonyms

FerPerales:
Asana ticket: https://app.asana.com/0/1203289004376659/1205605527270547/f

This PR:

* Add Synonyms model and admin section
* Connect db synonyms with ES synonym set through ES API
* Create ElasticsearchWrapper::Synonyms since this features are not fully supported yet
* Add rake task to create synonym set
* Update ES config for OccupationStandard to use dynamic synoyms set instead of fixed synonyms array

image


crystal-williams-brown (approved):
I'm not familiar with elastic search synonyms. This looks okay to me. I think it would be nice to have a seed file with some synonyms so that across deployment spaces there are a set that exist and impact queries. Currently it seems the synonyms would mostly be in production which means means queries elsewhere would produce different results.

.gitignore
```diff
@@ -44,3 +44,6 @@ spec/examples.txt
 
 # Brew files
 Brewfile.lock.json
+
+# kibana folder. asdf does not support kibana 8 so you will have to download it by hand
+kibana-*/*
```
littleforest, line 49: Instead of adding this to the repo's `.gitignore`, I would suggest putting this line in your `.git/info/exclude` file instead, which is like your own personal `.gitignore` file.

(You could also move your kibana setup to someplace like `/usr/local` if you might want to use it for other projects).

app/models/occupation_standard.rb
```diff
@@ -122,7 +132,7 @@ class OccupationStandard < ApplicationRecord
       indexes :rapids_code, type: :text, analyzer: :autocomplete, search_analyzer: :autocomplete_search
       indexes :state, type: :text, analyzer: :keyword
       indexes :state_id, type: :keyword
-      indexes :title, type: :text, analyzer: :rebuilt_english
+      indexes :title, type: :text, analyzer: :rebuilt_english, search_analyzer: :dynamic_synonym
```
littleforest, line 135: Since `rebuilt_english` isn't customized any longer, I think we should be able to remove the definition above, and just use `analyzer: :english` here.

lib/elasticsearch_wrapper/synonyms.rb
```diff
@@ -0,0 +1,45 @@
+module ElasticsearchWrapper
+  class Synonyms
+    SYNONYM_SET_NAME = "dynamic_synonyms_#{Rails.env}"
+    def self.client
+      @@client ||= Elasticsearch::Client.new
```
littleforest, line 5: This is not going to connect to the prod server correctly -- can we use `Elasticsearch::Model.client` which is set [here](https://github.com/ApprenticeshipStandardsDotOrg/ApprenticeshipStandardsDotOrg/blob/main/config/initializers/elasticsearch.rb#L15)?

lib/tasks/create_synonym_set.rake
```diff
@@ -0,0 +1,12 @@
...
+namespace :elasticsearch do
+  desc "Create Synonym Set on ES"
+  task create_synonym_set: :environment do
+    synonym = Synonym.create!(word: "UX", synonyms: "User experience")
+    ElasticsearchWrapper::Synonyms.create_set(
+      value: synonym.to_elasticsearch_value,
+      rule_id: synonym.id
+    )
+  end
+end
```
littleforest, line 12: Can we move this to an After Party task?

spec/lib/elasticsearch_wrapper/synonyms_spec.rb
```diff
@@ -0,0 +1,217 @@
...
+          value: value
+        )
+      ).to be true
+    end
+
+    it "throws an error when synonym set name does not exist" do
+      value = "Software Engineer, Developer"
+      id = "1"
+
+      stub_const("ElasticsearchWrapper::Synonyms::SYNONYM_SET_NAME", "unexisting")
```
littleforest, line 64:
```suggestion
      stub_const("ElasticsearchWrapper::Synonyms::SYNONYM_SET_NAME", "nonexistent")
```

spec/lib/elasticsearch_wrapper/synonyms_spec.rb
littleforest, line 1: Is there any reason why we can't write specs that actually work against the ES server instead of stubbing the responses?

littleforest (commented):
@FerPerales this is awesome! I'm wondering about the data structure, and if we need to worry about duplicates as we add more records. For example, someone could add:

```
word: developer
synonyms: dev, software engineer
```

and also:

```
word: software engineer
synonyms: dev, developer
```

right? When these get sent to ES, they just get sent as one list, not separated out as word/synonym, so I was wondering if there is a structure we could use that could more easily help us catch duplicates? Like a PG array or something, where if someone adds a new word, we can check it and see if it is already on a synonym list? I'm not sure how many synonyms we will actually have, so maybe it is not really a big deal to worry about dups.

FerPerales:
> @FerPerales this is awesome! I'm wondering about the data structure, and if we need to worry about duplicates as we add more records. For example, someone could add:
> 
> ```
> word: developer
> synonyms: dev, software engineer
> ```
> 
> and also:
> 
> ```
> word: software engineer
> synonyms: dev, developer
> ```
> 
> right? When these get sent to ES, they just get sent as one list, not separated out as word/synonym, so I was wondering if there is a structure we could use that could more easily help us catch duplicates? Like a PG array or something, where if someone adds a new word, we can check it and see if it is already on a synonym list? I'm not sure how many synonyms we will actually have, so maybe it is not really a big deal to worry about dups.

I'll ask Patrick about this during the meeting. Implementing this chance will add a couple of hours to this feature. Even though we have the time to do it, want to make sure if it's worth the added complexity. If we will have a very small list, it may not be

spec/lib/elasticsearch_wrapper/synonyms_spec.rb
crystal-williams-brown: ES tends to cause flaky specs due to the low amount of records in specs. I would stub if a live response isn't necessary.

FerPerales: @littleforest +1 to @crystal-williams-brown. I don't want to do live request because I don't want to test the elasticsearch-api, I only want to make sure that the wrapper will behave correctly give the responses from the API. Also, I hope I can remove the wrapper once the official gem supports synonyms in the official release.

littleforest (approved):
🚢

PR #452, opened 2023-10-11
Format onet code when importing

jenlu33:
Asana ticket: [https://app.asana.com/0/1203289004376659/1204540724864282/f](https://app.asana.com/0/1203289004376659/1204540724864282/f)

I updated the Onet scraper so that if there is a code that is formatted incorrectly like "47.3011.00", then it will be formatted correctly to "47-3011.00".


littleforest (approved):
Nice!

PR #473, opened 2023-10-18
add instructions for turning off xpath security

crystal-williams-brown:
add instructions for turning off xpath security so that elasticsearch 8 will work with our current setup


littleforest (commented):
@crystal-williams-brown can you move this to the #elasticsearch-setup area of the README?

littleforest (approved):
Perfect, thanks!

PR #478, opened 2023-10-23
Set up Rack Attack

littleforest:
[Asana ticket](https://app.asana.com/0/1203289004376659/1205550593976927/f)

This sets up the Rack Attack gem which is used to throttle excessive requests.

This also turns on paranoid mode for Devise for password resets, meaning the app will not tell you if the email entered was found in the database.


FerPerales (approved):
🚢

PR #479, opened 2023-10-23
Remove flipper flag for updated home

crystal-williams-brown:
I removed the conditional for displaying the updated home page so that it will always show the updated home page. I removed the flipper from tests as well.

[Asana ticket](https://app.asana.com/0/1203289004376659/1205773488666244/f)

### DIFF
--- a/app/views/pages/home.html.erb
+++ b/app/views/pages/home.html.erb
@@ -27,7 +27,7 @@
         <% end %>
     </div>
     <div class="text-center space-y-8 md:grid md:grid-cols-2 lg:grid-cols-3 xl:grid-cols-4 md:gap-8 xl:gap-8 md:space-y-0">
-      <% @popular_states.each do |state| %>
+      <% @popular_states&.each do |state| %>
         <%= link_to occupation_standards_path(q: "state:#{state.abbreviation}"), id: state.name.downcase do %>
           <div class="p-6 hover:bg-white rounded-lg hover:shadow-xl">
             <div class="mx-auto flex justify-center items-center mb-4 w-20 h-20 rounded-full bg-accent-200">
--- a/app/views/pages/home.html.erb
+++ b/app/views/pages/home.html.erb
@@ -27,7 +27,7 @@
         <% end %>
     </div>
     <div class="text-center space-y-8 md:grid md:grid-cols-2 lg:grid-cols-3 xl:grid-cols-4 md:gap-8 xl:gap-8 md:space-y-0">
-      <% @popular_states&.each do |state| %>
+      <% @popular_states.each do |state| %>
         <%= link_to occupation_standards_path(q: "state:#{state.abbreviation}"), id: state.name.downcase do %>
           <div class="p-6 hover:bg-white rounded-lg hover:shadow-xl">
             <div class="mx-auto flex justify-center items-center mb-4 w-20 h-20 rounded-full bg-accent-200">
--- a/app/views/pages/home.html.erb
+++ b/app/views/pages/home.html.erb
@@ -17,44 +17,18 @@
     </div>
   </div>
 </section>
-<% if Flipper.enabled?(:updated_home) %>
-  <section aria-labelledby="section-title" class="bg-accent-300">
-    <div class="py-8 px-4 mx-auto max-w-screen-xl sm:py-16 lg:px-6">
-      <div class="mb-8 lg:mb-16 flex justify-between">
-        <h2 class="section-title mb-4 text-4xl tracking-tight font-bold text-gray-900">Active States</h2>
-          <%= link_to states_path, class: "font-bold text-accent-50 sm:text-lg uppercase" do %>
-            See All
-            <svg aria-label="right arrow" class="inline mb-1" xmlns="http://www.w3.org/2000/svg" width="24" height="24" fill="none" stroke="currentColor" stroke-linecap="round" stroke-linejoin="round" stroke-width="2" class="feather feather-arrow-right"><path d="M5 12h14M12 5l7 7-7 7" /></svg>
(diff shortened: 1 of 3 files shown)


app/views/pages/home.html.erb
```diff
@@ -17,44 +17,18 @@
...
-  <!-- Popular types -->
-  <section aria-labelledby="section-title" class="bg-accent-300">
-    <div class="py-8 px-4 mx-auto max-w-screen-xl sm:py-16 lg:px-6">
-      <div class="mb-8 lg:mb-16 flex justify-between">
-        <h2 class="section-title mb-4 text-4xl tracking-tight font-bold text-gray-900">Popular Types</h2>
-      </div>
-      <div class="text-center space-y-8 md:grid md:grid-cols-2 lg:grid-cols-3 xl:grid-cols-4 md:gap-8 xl:gap-8 md:space-y-0">
-        <%= link_to occupation_standards_path(national_standard_type: {guideline_standard: 1}) do %>
+    <div class="text-center space-y-8 md:grid md:grid-cols-2 lg:grid-cols-3 xl:grid-cols-4 md:gap-8 xl:gap-8 md:space-y-0">
+      <% @popular_states&.each do |state| %>
```
littleforest, line 30: I don't think we should need `&.` here since `@popular_states` should be an Active Record relation (even if it's an empty one).

spec/system/pages/home_spec.rb
```diff
@@ -30,9 +30,6 @@
 
         visit home_page_path
 
-        expect(page).to have_text "Featured"
-
-        expect(page).to have_link "National Guidelines", href: occupation_standards_path(national_standard_type: {guideline_standard: 1})
```
littleforest, line 35: Can we at least check that this "National Guidelines" text exists, even if we don't test for the link anymore?

spec/system/pages/home_spec.rb
```diff
@@ -55,7 +52,6 @@
 
         visit home_page_path
 
-        expect(page).to have_link "Occupational Frameworks", href: occupation_standards_path(national_standard_type: {occupational_framework: 1})
```
littleforest, line 58: Similar here, can we still check for the "Occupational Frameworks" text?

spec/system/pages/home_spec.rb
```diff
@@ -68,22 +64,18 @@
...
         ra = create(:registration_agency, state: wa)
+        ca_ra = create(:registration_agency, state: ca)
         mechanic = create(:occupation_standard, :with_work_processes, :with_data_import, registration_agency: ra, title: "Mechanic")
         hr = create(:occupation_standard, :with_work_processes, :with_data_import, registration_agency: ra, title: "HR")
-        create(:occupation_standard, :with_work_processes, title: "Pipe Fitter")
+        create(:occupation_standard, :with_work_processes, registration_agency: ca_ra, title: "Pipe Fitter")
 
         visit home_page_path
 
-        expect(page).to have_link "Washington", href: occupation_standards_path(state_id: wa.id)
```
littleforest, line 86: And here, can we add in a check for "Washington" text.

spec/system/pages/home_spec.rb
```diff
@@ -95,125 +87,21 @@
...
-      it "displays Healthcare Support industry box" do
-        allow(State).to receive(:find_by).and_return(build_stubbed(:state))
-
-        mechanic = create(:occupation_standard, :with_work_processes, :with_data_import, onet_code: "31-1234", title: "Mechanic")
-        hr = create(:occupation_standard, :with_work_processes, :with_data_import, onet_code: "31-5678", title: "HR")
-        create(:occupation_standard, :with_work_processes, onet_code: "35-1234", title: "Pipe Fitter")
-
-        visit home_page_path
-
-        expect(page).to have_link "Healthcare Support", href: occupation_standards_path(q: "31-")
```
littleforest, line 209: Can we check for "Healthcare Support" text?

littleforest (commented):
Left some comments about checking for some text, didn't carry those comments down into the ES section, but similar request for those specs as well.

app/views/pages/home.html.erb
crystal-williams-brown: This was because a test complained about popular_states being nil. I don't recall which test it was.

spec/system/pages/home_spec.rb
crystal-williams-brown: We are clicking on "National Guidelines" to get to the page where we check the link. Same is true for the other sections (State, Occupational Frameworks, etc)

spec/system/pages/home_spec.rb
crystal-williams-brown: For this text check are you wanting me to add that expectation for the new view we've navigated to (not the home page)?

crystal-williams-brown: Is the "Healthcare Support" text supposed to be on the search results page?

app/views/pages/home.html.erb
littleforest: You will need to assign `:popular_states` in [this view spec](https://github.com/ApprenticeshipStandardsDotOrg/ApprenticeshipStandardsDotOrg/blob/main/spec/views/pages/home_spec.rb) and then you can revert this change.

spec/system/pages/home_spec.rb
littleforest: Oh you're right, sorry!

app/views/pages/home.html.erb
crystal-williams-brown: I've updated the test and reverted the change

littleforest (approved):
🚢

PR #480, opened 2023-10-23
Set Onet to current version when importing occupations

littleforest:
This sets the search on onet codes to the current version when importing occupations.


jenlu33 (approved):

littleforest:
> Could it be a scope like Onet.current_version.find_by. Not sure which way performs better but I imagine needing to filter by current version in other circumstances.

Ooo I like that. I will make a separate ticket to make that change everywhere.

PR #486, opened 2023-10-27
add task to cleanup onet codes

crystal-williams-brown:
I may need to manually fix the nils and the record where the `onet_code` is "onet". I will first check if I can programmatically fix those as well. For now this fixes the other onet_codes that aren't properly formatted.

[Asana ticket](https://app.asana.com/0/1203289004376659/1205773488666252/f)

### DIFF
--- a/app/models/occupation_standard.rb
+++ b/app/models/occupation_standard.rb
@@ -385,7 +385,8 @@ class OccupationStandard < ApplicationRecord
       return nil if onet_code == "onet"
 
       base_onet = base_onet.delete("^0-9")
-      base_onet << "00" if base_onet.length == 6
+      return onet_code if base_onet.length < 8
+      
       base_onet = base_onet.insert(2, "-").insert(7, ".")
     end
 
--- a/app/models/occupation_standard.rb
+++ b/app/models/occupation_standard.rb
@@ -377,6 +377,20 @@ class OccupationStandard < ApplicationRecord
     state&.id
   end
 
+  def clean_onet_code
+    # correct format 12-3456.07
+    base_onet = onet_code
+
+    if onet_code.present?
+      return nil if onet_code == "onet"
+
+      base_onet = base_onet.strip.gsub!(/[^0-9A-Za-z]/, "")
+      base_onet = base_onet.insert(2, "-").insert(7, ".")
+    end
+
+    base_onet
+  end
+
   private
 
   def national?
--- a/app/models/occupation_standard.rb
+++ b/app/models/occupation_standard.rb
@@ -384,7 +384,8 @@ class OccupationStandard < ApplicationRecord
     if onet_code.present?
       return nil if onet_code == "onet"
 
-      base_onet = base_onet.strip.gsub!(/[^0-9A-Za-z]/, "")
+      base_onet = base_onet.delete("^0-9")
+      base_onet << "00" if base_onet.length == 6
       base_onet = base_onet.insert(2, "-").insert(7, ".")
     end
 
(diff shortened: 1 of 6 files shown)


crystal-williams-brown:
I'm not sure if I'm meant to change the onet_codes on the occupation standards. Meaning change the format for the ones that aren't `nil`

littleforest:
> I'm not sure if I'm meant to change the onet_codes on the occupation standards. Meaning change the format for the ones that aren't nil

@crystal-williams-brown yes, please write a script to clean up the production data.

spec/models/occupation_standard_spec.rb
```diff
@@ -792,4 +792,45 @@
...
+      expect(occupation_standard.clean_onet_code).to be_nil
+    end
+
+    it "does not modify the onet_code column" do
+      occupation_standard = create(:occupation_standard, onet_code: "17-3024.00   ")
+
+      expect do
+        occupation_standard.clean_onet_code
+        occupation_standard.reload
+      end.not_to change(occupation_standard, :onet_code)
```
littleforest, line 833: Why don't we want the `onet_code` changed? We would like the extra whitespace stripped.

spec/models/occupation_standard_spec.rb
littleforest, line 1: There are more examples to add to your specs. For example, the ones like: `39-9011.00T` and the ones like: `51-4011`.

littleforest (changes requested):

spec/models/occupation_standard_spec.rb
crystal-williams-brown: This is a formatter for the after party task. In the after party task the onet code is changed

spec/models/occupation_standard_spec.rb
crystal-williams-brown: With all the variations I think today I will have to manually fix all the codes in production. It seems like a script is not ideal for this.

app/models/occupation_standard.rb
```diff
@@ -377,6 +377,20 @@ def state_id
...
   end
 
+  def clean_onet_code
+    # correct format 12-3456.07
+    base_onet = onet_code
+
+    if onet_code.present?
+      return nil if onet_code == "onet"
+
+      base_onet = base_onet.strip.gsub!(/[^0-9A-Za-z]/, "")
```
littleforest, line 387:
```suggestion
      base_onet = base_onet.strip.gsub(/[^0-9A-Za-z]/, "")
```
The bang replaces in place, but since you are assigning back to `base_onet`, you don't want the bang.

spec/models/occupation_standard_spec.rb
crystal-williams-brown: I added some format fixes for outliers and skips for the too short codes. I've also added more tests.

app/models/occupation_standard.rb
```diff
@@ -377,6 +377,22 @@ def state_id
...
 
+  def clean_onet_code
+    # correct format 12-3456.07
+    base_onet = onet_code
+
+    if onet_code.present?
+      return nil if onet_code == "onet"
+
+      base_onet = base_onet.delete("^0-9")
+      return onet_code if base_onet.length < 8
```
littleforest, line 388:
```suggestion
      return onet_code if base_onet.length < 7
```

You're trying to skip the ones like "12-3456", so wouldn't it be less than 7, since at this line, base_onet will look like: "123456" ?

crystal-williams-brown: The expected length is 8. I don't know if the ones that are 7 need extra care too. 12-3456 is 2 digits shorter than the other onets.

littleforest (approved):
:shipit:

PR #496, opened 2023-11-07
Add fact sheets page and link in footer

jenlu33:
[Asana ticket](https://app.asana.com/0/1203289004376659/1205901391498411/f)

Just waiting on the links to be separated out to be two separate links. Right now both links have the same url


littleforest (approved):
Looks good to me once you get the new link!

littleforest (approved):

PR #509, opened 2023-11-16
Freeze time

FerPerales:
[Asana ticket](https://app.asana.com/0/1203289004376659/1205931295164762/f)

Freeze time. Ran this locally 100 times and it seems is fixed now

image


littleforest (approved):
🎉

PR #535, opened 2023-12-04
Add recently added section to home page

FerPerales:
[Asana ticket](https://app.asana.com/0/1203289004376659/1205982289180736/f)

This add a Recently Added section to home page with the last 4 most recent Occupation Standards in the system.
This also adds a link to Occupation Standards index page with a filter to order results by creation date dismissing the score

Recently Added section:

image


app/models/occupation_standard.rb
```diff
@@ -218,6 +219,8 @@ def as_indexed_json(_ = {})
 
   scope :with_work_processes, -> { joins(:work_processes).distinct }
 
+  scope :recently_added, -> { order(created_at: :desc).limit(MAX_RECENTLY_ADDED_OCCUPATIONS_TO_DISPLAY) }
```
littleforest, line 222: @FerPerales I think we still need to hook into ES here, or otherwise skip occupation standards that have no work processes ([per this constraint](https://github.com/ApprenticeshipStandardsDotOrg/ApprenticeshipStandardsDotOrg/blob/main/app/queries/occupation_standard_elasticsearch_query.rb#L29-L31)) if we want to have a postgres query here. It seems a little overkill to go to ES for this; on the other hand, it will make sure our queries are consistent if we add other constraints. I will defer to you.

jenlu33 (approved):

app/models/occupation_standard.rb
FerPerales: @littleforest I updated the scope to exclude occupation standards without work processes using [associated](https://api.rubyonrails.org/classes/ActiveRecord/QueryMethods/WhereChain.html#method-i-associated)

app/models/occupation_standard.rb
```diff
@@ -218,6 +219,8 @@ def as_indexed_json(_ = {})
 
   scope :with_work_processes, -> { joins(:work_processes).distinct }
 
+  scope :recently_added, -> { where.associated(:work_processes).order(created_at: :desc).limit(MAX_RECENTLY_ADDED_OCCUPATIONS_TO_DISPLAY) }
```
littleforest, line 222: cool, I did not know about this new Rails 7 `associated` method!

spec/models/occupation_standard_spec.rb
```diff
@@ -199,6 +199,29 @@
     end
   end
 
+  describe ".recently_added" do
+    it "excludes occupations without work processes" do
+      create_list(:occupation_standard, 4)
```
littleforest, line 204: Can we just create 2 objects here instead of 4? It will serve the same purpose and reduces DB insert count.

spec/models/occupation_standard_spec.rb
```diff
@@ -199,6 +199,29 @@
...
+      occupation_with_work_proccesses = create(:occupation_standard, :with_work_processes)
+
+      expect(described_class.recently_added).to match [occupation_with_work_proccesses]
+    end
+
+    it "returns up to MAX_RECENTLY_ADDED_OCCUPATIONS_TO_DISPLAY" do
+      create_list(:occupation_standard, 5, :with_work_processes)
+
+      expect(described_class.recently_added.count).to eq described_class::MAX_RECENTLY_ADDED_OCCUPATIONS_TO_DISPLAY
```
littleforest, line 213: What do you think about stubbing this to reduce DB count and not making the spec dependent on the constant?

```
stub_const("OccupationStandard::MAX_RECENTLY_ADDED_OCCUPATIONS_TO_DISPLAY", 2)
create_list(:occupation_standard, 3, :with_work_processes)

expect(described_class.recently_added.count).to eq 2
```

littleforest (approved):

PR #553, opened 2023-12-19
Fix N+1 issue on occupation_standards#index page

littleforest:
The inner hits method collects occupation standard records into an array, but then we end up with some N+1 queries. This re-collects the occupation standards in an Active Record relation so we can take advantage of `includes`.


FerPerales (approved):
Good catch! :shipit:

PR #557, opened 2023-12-21
Export and saves attachments out of docx file

Joranhezon:
Adds the ability to export docx files attachments automatically, without needing someone to do so manually.

Creates a service that exports and saves files attached to docx files. To do so, we are passing a SourceFile object and a file blob for the service. With those, we are able to transform the docx file into a zip file, which is necessary for getting the attached files out. We then unzip the file, getting a list of attached files and their respective paths. Finally, we create a StandardImport based on the SourceFile passed to the service, and attach each of the extracted files to it. As a clean up stage, we unlink the tempfile created for the zip, and delete the extracted files.

[Asana ticket](https://app.asana.com/0/1203289004376659/1206122919120651/f)


app/services/export_file_attachments.rb
```diff
@@ -0,0 +1,68 @@
+class ExportFileAttachments
+  attr_reader :source_file, :blob
+
+  TEMP_FILE_PATH = "foo.zip"
+  BULLETIN_LIST_URL = "https://www.apprenticeship.gov/about-us/legislation-regulations-guidance/bulletins/export?search=&category%5B0%5D=National%20Guideline%20Standards&category%5B1%5D=National%20Program%20Standards&category%5B2%5D=Occupations&page&_format=csv"
```
Joranhezon, line 5: Is this the source url for the StandardImport created here? I'm confused.

app/services/export_file_attachments.rb
```diff
@@ -0,0 +1,68 @@
+class ExportFileAttachments
+  attr_reader :source_file, :blob
+
+  TEMP_FILE_PATH = "foo.zip"
```
Joranhezon, line 4: Should we store it temporarily somewhere else?

app/services/export_file_attachments.rb
Joranhezon, line 5: Is this the source url for the StandardImport created here? I think I'm missing some context.

app/services/export_file_attachments.rb
```diff
@@ -0,0 +1,68 @@
...
+  end
+
+  def save_attachments_to_db(file_names)
+    standards_import = StandardsImport.where(
+      name: "Source File #{source_file.id}",
+      organization: nil
+    ).first_or_initialize(
+      notes: "Extracted from Apprenticeship Bulletings",
+      public_document: true,
+      source_url: BULLETIN_LIST_URL
```
Joranhezon, line 50: Same question as above

app/services/export_file_attachments.rb
```diff
@@ -0,0 +1,68 @@
...
+    @source_file = source_file
+    @blob = blob
+  end
+
+  def call
+    zip_file = create_zip_version_of_source_file
+    file_names = unzip_attachments_and_list_file_names
+    save_attachments_to_db(file_names)
+    delete_extracted_files(file_names, zip_file)
+  end
```
Joranhezon, line 17: How are we handling exceptions in the project? I wanted to add a generic rescue here, but don't know what would do in the treatment stage.

Joranhezon:
Where am I supposed to add this service? Is it in the `Scraper::AppreticeshipBulletinsJob`? I noticed we're already creating a StandardImport there, so I'm confused as to how move forward with this.

app/services/export_file_attachments.rb
littleforest: This is the URL that we use to scrape the bulletins from.

littleforest: You shouldn't need it in this new service.

littleforest (commented):
@Joranhezon this looks really good! Can you call this new service after [this line](https://github.com/ApprenticeshipStandardsDotOrg/ApprenticeshipStandardsDotOrg/blob/main/app/jobs/scraper/appreticeship_bulletins_job.rb#L31). Though now I am wondering if this service should actually replace the guts of the `Scraper::AppreticeshipBulletinsJob` (maybe we don't even need to save those files to a StandardsImport record, and we should just extract them immediately. I will double check).

app/services/export_file_attachments.rb
```diff
@@ -0,0 +1,69 @@
...
+    zip_file = create_zip_version_of_source_file
+    file_names = unzip_attachments_and_list_file_names(zip_file)
+    save_attachments_to_db(file_names)
+    delete_extracted_files(file_names, zip_file)
+  end
+
+  private
+
+  def create_zip_version_of_source_file
+    zip_file = Tempfile.open(TEMP_FILE_PATH, encoding: "ascii-8bit")
```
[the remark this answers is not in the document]
Joranhezon: Unf yes, when I was trying to write to a tempfile in UTC-8 I was getting an error.

app/services/export_file_attachments.rb
```diff
@@ -0,0 +1,69 @@
...
+  def unzip_attachments_and_list_file_names(zip_file)
+    file_names = []
+
+    Zip::File.open(zip_file) do |zip_file|
+      zip_file.each do |entry|
+        next unless entry.name.include?("docx") || entry.name.include?(".bin")
+        entry.name.sub!(".bin", ".pdf") if entry.name.include?(".bin")
+
+        file_path = "#{Rails.root}/tmp/#{File.basename(entry.name)}"
+        entry.extract(@entry_path = file_path)
```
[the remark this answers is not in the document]
Joranhezon: This is to set the path in which we're extracting the files. I put in the tmp folder, cause otherwise we would need to create a word/embeddings folder, as that would be the default here.

spec/services/export_file_attachments_spec.rb
```diff
@@ -0,0 +1,15 @@
...
+  describe "#call" do
+    it "creates a StandardImport and 6 files attached to it" do
+      create(:standards_import, :with_docx_file_with_attachments)
+      source_file = SourceFile.last
+      blob = source_file.active_storage_attachment.blob
+
+      expect { described_class.new(source_file, blob).call }
+        .to change { StandardsImport.count }.by(1)
+        .and change { StandardsImport.last.files.count }.by(6)
+    end
```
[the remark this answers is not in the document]
Joranhezon: Great idea!

app/jobs/scraper/export_file_attachments_job.rb
```diff
@@ -0,0 +1,67 @@
...
+  def create_zip_version_of_source_file(source_file)
+    zip_file = Tempfile.open(TEMP_FILE_PATH, encoding: "ascii-8bit")
+    source_file.active_storage_attachment.blob.download { |chunk| zip_file.write(chunk) }
+    zip_file
+  end
+
+  def unzip_attachments_and_list_file_names(zip_file)
+    file_names = []
+
+    if !File.zero?(zip_file)
```
FerPerales, line 24: @Joranhezon what are your thoughts on replacing this with `unless File.empty?(zip_file)`

app/jobs/scraper/export_file_attachments_job.rb
```diff
@@ -0,0 +1,67 @@
...
+      Zip::File.open(zip_file) do |zip_file|
+        zip_file.each do |entry|
+          next unless entry.name.include?("docx") || entry.name.include?(".bin")
```
FerPerales, line 27: We probably want to make sure that both docx and bin are present in the filename as the extension and not just in the middle (I would be surprise if that's the case, but who knows). We can use `end_with?` to guarantee this

app/jobs/scraper/apprenticeship_bulletins_job.rb
```diff
@@ -1,4 +1,4 @@
-class Scraper::AppreticeshipBulletinsJob < ApplicationJob
+class Scraper::ApprenticeshipBulletinsJob < ApplicationJob
```
FerPerales, line 1: @Joranhezon Good catch!

FerPerales (approved):
Looks good! Left a couple of comments.

app/jobs/scraper/export_file_attachments_job.rb
Joranhezon: I like it! Good catch, thanks.

app/jobs/scraper/export_file_attachments_job.rb
Joranhezon: Ooo nice!

PR #561, opened 2023-12-27
Update document when importing completed

FerPerales:
[Asana ticket](https://app.asana.com/0/1203289004376659/1206163711596811/f)

When importing an Occupation Standard, we add the entry to ElasticSearch the first time we create the occupation, but we never updated.

This PR add a call to update_document to make sure we reindex the Occupation Standard with the updated information


Joranhezon (approved):
LGTM!

PR #565, opened 2023-12-28
Install libreoffice

FerPerales:
libreoffice allows to convert docx to pdf via a command-line utility

```
soffice  --headless --convert-to pdf test.docx --outdir .
```

Once libreoffice is installed, we can do some testing to integrate the `soffice` command-line conversion with Ruby


littleforest (approved):
Is this something that we should also include instructions in the README for how to install locally?

FerPerales:
@littleforest Definitely. Just updated the README with instructions. Thanks!

PR #568, opened 2024-01-02
Chatgpt work process

Joranhezon:
Add a script that uses chatgpt to extract work process and return a json with the data. The idea is for this to be used in the ProcessDataImportJob behind a feature flag, so we can start filling the database in an automated manner, instead of using a spreadsheet.

[Asana ticket](https://app.asana.com/0/1203289004376659/1205945047456744)


app/jobs/work_process_extractor_job.rb
```diff
@@ -0,0 +1,13 @@
+class WorkProcessExtractorJob < ApplicationJob
+  queue_as :default
+
+  def perform()
+    file = '/Users/nina/Documents/ApprenticeshipStandardsDotOrg/tmp/test4.pdf'
+    reader = PDF::Reader.new(file)
+    text = reader.pages.map{|page| page.text}
```
Joranhezon, line 7: Where are we getting the file from?

littleforest: These would be all of the ActiveStorage::Attachment files that are linked to the SourceFile records.

FerPerales:
@rdnewmanbot yes. Latest branches related to chatGPT work are based on this one

PR #576, opened 2024-01-05
Add service to convert word documents to PDF

FerPerales:
[Asana ticket](https://app.asana.com/0/1203289004376659/1206122919120653/f)

This services is a wrapper to interact with LibreOffice conversion tool.

### DIFF
--- a/spec/services/doc_to_pdf_converter_spec.rb
+++ b/spec/services/doc_to_pdf_converter_spec.rb
@@ -1,0 +1,32 @@
+require "rails_helper"
+
+RSpec.describe DocToPdfConverter do
+  describe "#call" do
+    it "raises error if libreoffice not installed" do
+      source_file = build(:source_file)
+      allow(Kernel).to receive(:system).with("command -v soffice").and_return false
+
+      expect { DocToPdfConverter.call(source_file) }.to raise_exception DocToPdfConverter::DependencyNotFoundError
+    end
+
+    it "raises error if conversion failed" do
+      source_file = build(:source_file)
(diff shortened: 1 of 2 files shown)


jenlu33 (approved):

spec/services/doc_to_pdf_converter_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+    it "raises error if libreoffice not installed" do
+      source_file = build(:source_file)
+      allow(Kernel).to receive(:system).with("command -v soffice").and_return false
+
+      expect { DocToPdfConverter.call(source_file) }.to raise_exception DocToPdfConverter::DependencyNotFoundError
+    end
+
+    it "raises error if conversion failed" do
+      source_file = build(:source_file)
+      FileUtils.remove_dir(DocToPdfConverter::TMP_DIRECTORY, true) if File.exist?(DocToPdfConverter::TMP_DIRECTORY)
```
littleforest, line 14: Wondering if we should be consistent between using `FileUtils` vs `Dir`?

littleforest (approved):

spec/services/doc_to_pdf_converter_spec.rb
FerPerales: @littleforest `Dir.rm_dr` raises an error if the folder is not empty and there's no way to force, while `FileUtils.remove_dir` allows it.

littleforest: Oh nice, good to know.

PR #577, opened 2024-01-05
Extract WorkProcess data using chatgpt

Joranhezon:
Add a job that uses chatgpt to extract work process and fill the database with data. The idea is for this to be used in the ProcessDataImportJob behind a feature flag, so we can start filling the database in an automated manner, instead of using a spreadsheet.

[Asana ticket](https://app.asana.com/0/1203289004376659/1205945047456744)


app/jobs/work_process_extractor_job.rb
```diff
@@ -0,0 +1,39 @@
+class WorkProcessExtractorJob < ApplicationJob
+  queue_as :default
+
+  def perform(occupation_standard, competency_data)
+    file = "/Users/nina/Documents/ApprenticeshipStandardsDotOrg/tmp/test3.pdf"
```
Joranhezon, line 5: Where are we getting this file from?

rdnewmanbot: (see https://github.com/ApprenticeshipStandardsDotOrg/ApprenticeshipStandardsDotOrg/pull/568#discussion_r1443391632)

app/services/import_via_chatgpt/work_processes.rb
```diff
@@ -0,0 +1,25 @@
...
+    def import
+      work_process = WorkProcess.find_or_initialize_by(
+        occupation_standard:,
+        title: description.presence
+      )
+      work_process.update!(description:, minimum_hours:, maximum_hours:)
+      work_process
+    end
+
+    private
```
rdnewmanbot, line 21: looks like I could drop this private section now

PR #595, opened 2024-01-23
The system makes clear the list of source files that need to be redacted

FerPerales:
[Asana ticket](https://app.asana.com/0/1203289004376659/1206305102472868/f)

* Add a link in Source File admin dashboard to filter sources files that have a PDF file, are completed, and don't have a redacted file version yet

image

### DIFF
--- a/app/models/source_file.rb
+++ b/app/models/source_file.rb
@@ -8,6 +8,8 @@ class SourceFile < ApplicationRecord
   enum :status, [:pending, :completed, :needs_support, :needs_human_review]
   enum courtesy_notification: [:not_required, :pending, :completed], _prefix: true
 
+  PDF_CONTENT_TYPE = "application/pdf"
+
   def filename
     active_storage_attachment.blob.filename
   end
@@ -43,7 +45,7 @@ class SourceFile < ApplicationRecord
   end
 
   def pdf?
-    active_storage_attachment.blob.content_type == "application/pdf"
+    active_storage_attachment.blob.content_type == PDF_CONTENT_TYPE
   end
 
   def redacted_source_file_url
@@ -59,6 +61,16 @@ class SourceFile < ApplicationRecord
       redacted_at: (
         start_time..end_time
       )
+    )
+  end
+
+  def self.pdf_attachment
+    includes(active_storage_attachment: :blob).where(
+      active_storage_attachment: {
+        active_storage_blobs: {
+          content_type: PDF_CONTENT_TYPE
+        }
+      }
     )
   end
 end
--- a/app/models/source_file.rb
+++ b/app/models/source_file.rb
@@ -81,4 +81,8 @@ class SourceFile < ApplicationRecord
       }
     )
   end
+
+  def self.ready_for_redaction
+    completed.not_redacted.pdf_attachment
+  end
 end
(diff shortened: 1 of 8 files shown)


spec/factories/source_files.rb
```diff
@@ -8,5 +8,37 @@
         active_storage_attachment: standards_import.files.first
       ).first
     end
+
+    trait :with_docx_attachment do
```
jutonz, line 12: ah awesome, these will be very handy for what I'm doing. Thanks very much!

app/views/admin/source_files/_index_header.html.erb
```diff
@@ -7,6 +7,22 @@
...
+    <% if params[:ready_for_redaction] %>
+      <%= link_to(
+            "Show all files",
+            admin_source_files_path,
+            class: "button"
+          ) %>
+  <% else %>
+    <%= link_to(
+          "Only show files ready for redaction",
+          admin_source_files_path(ready_for_redaction: true),
```
jutonz, line 20: What you have is totally fine but just FYI I found about this cool [`url_for`](https://apidock.com/rails/ActionView/RoutingUrlFor/url_for) helper recently and it's perfect for situations like this. It has a lot of different uses, but one of them is to just take the current URL and add or remove query params. So for example you could use it like this

```suggestion
          url_for(ready_for_redaction: true),
```

That way you don't have to repeat the path helper over and over when you just want to change the query params.

Again no need to change what you have here--I'm just sharing because I thought it was cool!

spec/factories/standards_imports.rb
```diff
@@ -8,6 +8,10 @@
       files { [Rack::Test::UploadedFile.new(Rails.root.join("spec", "fixtures", "files", "docx_file_attachments.docx"))] }
     end
 
+    trait :with_pdf_file_with_attachments do
```
jutonz, line 11: Can a PDF file have attachments in the same way a docx file can? If not how do you feel about this name instead? 
```suggestion
    trait :with_pdf_file do
```

jutonz (approved):
awesome, looks great! I'm not fully up to speed yet so maybe good to wait for someone else's review too, but to my eye this looks really good :)

app/views/admin/source_files/_index_header.html.erb
FerPerales: @jutonz Thanks great! Thanks for sharing! I'll apply this

spec/factories/standards_imports.rb
FerPerales: @jutonz good catch! PDF do not have any attachments

app/models/source_file.rb
```diff
@@ -61,4 +63,26 @@ def self.recently_redacted(start_time: Time.zone.yesterday.beginning_of_day, end
...
+    )
+  end
+
+  def self.not_redacted
+    includes(:redacted_source_file_attachment).where(
+      redacted_source_file_attachment: {
+        id: nil
+      }
+    )
+  end
```
littleforest, line 83: Can this method and the one above be private methods?

littleforest (approved):
Looks good to me, just one minor comment/question!

app/models/source_file.rb
FerPerales: @littleforest I tried this suggestion, but it's breaking the implementation. My guess is that combining scopes (`completed` from the enum) and private class methods does not work together well 🤔

PR #605, opened 2024-01-25
Add bin/elasticsearch script

jutonz:
Just a super tiny script to detect if ES is already running before attempting to start it. I've been running ES in docker but commenting out that line of `Procfile.dev` to avoid a conflict. Hopefully this simplifies things a bit for future devs!


littleforest (approved):
Nice! 🎉 I hit this all the time.

PR #606, opened 2024-01-25
Hide archived source files by default

jutonz:
On the admin source files index page, archived files are now hidden unless you explicitly search for them.

[Asana ticket](https://app.asana.com/0/1203289004376659/1206228142308915)

### DIFF
--- a/app/controllers/admin/source_files_controller.rb
+++ b/app/controllers/admin/source_files_controller.rb
@@ -13,6 +13,13 @@ module Admin
     end
 
     def filter_resources(resources, search_term:)
+      resources =
+        if Administrate::SourceFileSearch.user_is_filtering_by_status?(search_term)
+          resources
+        else
+          resources.not_archived
+        end
+
       Administrate::SourceFileSearch.new(
         resources,
         dashboard,
@@ -36,6 +43,14 @@ end
 
 module Administrate
   class SourceFileSearch < Search
+    def self.user_is_filtering_by_status?(term)
+      db_value_for_status(term).present?
+    end
+
+    def self.db_value_for_status(term)
+      SourceFile.statuses[term.parameterize(separator: "_")]
+    end
+
     private
 
     def query_template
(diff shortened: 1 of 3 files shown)


app/controllers/admin/source_files_controller.rb
```diff
@@ -13,6 +13,13 @@ def after_resource_updated_path(resource)
     end
 
     def filter_resources(resources, search_term:)
+      resources =
+        if Administrate::SourceFileSearch.user_is_filtering_by_status?(search_term)
```
jutonz, line 17: I would have preferred to have this conditional up in `scoped_resource` above, but that was a little tricky since we don't have access to the `search_term` param. I tried just using `params["search"]` but I guess administrate does some processing to the raw param value before passing it in here, so it seemed simpler to just adjust this method. I don't think this is super weird though since the method is called `filter_resources` and we are in fact filtering resources 😅

app/controllers/admin/source_files_controller.rb
```diff
@@ -45,14 +60,26 @@ def query_template
         .push("LOWER(standards_imports.organization) LIKE ?")
         .push("LOWER(users.name) LIKE ?")
         .push("source_files.public_document = ?")
-        .push("status = ?")
         .join(" OR ")
+        .then { add_status_query(_1) }
```
jutonz, line 64: This isn't a totally necessary refactor but I think addresses one bit of potential confusion with the query in combination with our "default scope" above. Without this, what ends up in the query if you're not searching for anything (or searching for something other than a status) is `OR status = NULL`. That's actually fine since there's a null constraint on `status` so it's not really possible for it to be null. But still it felt a little weird to say `WHERE status != 'archived' OR status = NULL`.

To deal with that I just conditionally add the `OR status = ?` part based on whether the user passed in a valid status search. That also required conditionally adding the value below, since otherwise the number of values passed in wouldn't match the number of `?`'s present in the query. What do y'all think?

littleforest: @jutonz I'm excited about this but haven't used `.then` before and am having trouble searching for it. Do you have a reference link?

littleforest: Oh nevermind, this is on the Array object! I was thinking it was on Active Record. Found the info.

littleforest (approved):
🎉

PR #626, opened 2024-02-07
Add RAPIDS::API service

FerPerales:

This is the first part of the feature. I decided to split this into several PRs.

* Add oauth2 gem 
* Create RAPIDS models for OccupationStandard and Organization
* Add RegistrationAgency.registration_agency_for_national_program 
* Add RAPIDS::API service to connect to RAPIDS API
* Create factories for API responses


app/models/rapids/occupation_standard.rb
```diff
@@ -0,0 +1,53 @@
...
+          parsed_sponsor_name = sponsor_number.match(/-?(?<state>[A-Z]{2})/)
+
+          state_abbreviation = parsed_sponsor_name && parsed_sponsor_name[:state]
+
+          state = State.find_by(
+            abbreviation: state_abbreviation
+          )
+
+          if state
+            registration_agency = state.registration_agencies.last
```
littleforest, line 41: Some states have both OA and SAA -- do we know which one we would want to grab in those cases?

littleforest (commented):
Looks good to me, just one question.

app/models/rapids/occupation_standard.rb
FerPerales: @littleforest didn't know that case was possible 😮  I'll ask Patrick

FerPerales: It must use OA. Code updated to reflect this

PR #635, opened 2024-02-12
Add libreoffice to final Docker image

littleforest:
[Asana ticket](https://app.asana.com/0/1203289004376659/1206305102472876/f)

The libreoffice package was only added to the build stage of the Dockerfile, so it was not present in the final image.


FerPerales:
@littleforest oof. Good catch! Thanks!

PR #640, opened 2024-02-13
The admin section has a filter for already redacted files. 

FerPerales:
[Asana ticket](https://app.asana.com/0/1203289004376659/1206423468333638/f)

This PR unifies the previous filter and add the new one under a dropdown

image


littleforest (approved):
Ooo nice. Maybe we can use a filter for showing public documents eventually as well.

PR #675, opened 2024-03-04
Use unique file path for the extracted documents

littleforest:
When extracting documents from bulletins, we are
using the same tmp file directory which is resulting
in collisions. Use a unique tmp file directory for each
source file that is being processed.

These changes should have been made initially
as part of #672.


rdnewmanbot (approved):
In general I prefer to move all/most of this logic into a new class (under services probably) and keep the job thin (like a controller), esp. once an instance variable becomes necessary, but should be safe. Still it's a bit easier to test things that way. 

Otherwise looks good and nice to not be passing `source_file` around as an argument everywhere.

FerPerales (approved):

PR #683, opened 2024-03-11
Create import data from RAPIDS job

FerPerales:

This PR adds first iteration of `ImportDataFromRAPIDSJob`, capable of importing both `OccupationStandard` and `WorkProcess`. I'm going to split this ticket this the scope is larger than expected


app/jobs/import_data_from_rapids_job.rb
```diff
@@ -0,0 +1,51 @@
...
+
+    loop do
+      Rails.logger.info "Processing batchSize: #{PER_PAGE_SIZE}, startIndex: #{start_index}"
+      response = service.work_processes(
+        batchSize: PER_PAGE_SIZE,
+        startIndex: start_index
+      )
+      parsed_response = response.parsed
+      occupation_standards = process_api_response(parsed_response)
+      occupation_standards.map(&:save)
```
littleforest, line 18: Do we need to worry about not doing any error checking on `save` here? This is going to silently fail if we are missing any validations.

app/jobs/import_data_from_rapids_job.rb
```diff
@@ -0,0 +1,51 @@
...
+      total_records = parsed_response["totalCount"]
+      start_index += PER_PAGE_SIZE
+
+      break if start_index > total_records
```
littleforest, line 22: What about using a `while` loop instead of breaking out of this loop?

```
while start_index <= total_records
  ...
  start_index += PER_PAGE_SIZE
end
```

spec/models/rapids/competency_spec.rb
```diff
@@ -0,0 +1,13 @@
+require "rails_helper"
+
+RSpec.describe RAPIDS::Competency, type: :model do
+  describe ".initialize_from_response" do
+    it "returns organization with correct data" do
+      competency_response = create(:rapids_api_competency, title: "Competency #1")
+
+      competency = RAPIDS::Competency.initialize_from_response(competency_response)
```
littleforest, line 8:
```suggestion
      competency = described_class.initialize_from_response(competency_response)
```

spec/models/rapids/work_process_spec.rb
```diff
@@ -0,0 +1,20 @@
...
+  describe ".initialize_from_response" do
+    it "returns work process with correct data" do
+      work_process_response = create(
+        :rapids_api_detailed_work_activity,
+        title: "Sample work process",
+        minHours: 20,
+        maxHours: 200
+      )
+
+      work_process = RAPIDS::WorkProcess.initialize_from_response(work_process_response)
```
littleforest, line 13:
```suggestion
      work_process = described_class.initialize_from_response(work_process_response)
```

app/jobs/import_data_from_rapids_job.rb
FerPerales: @littleforest staging server has a lot of invalid data and it may be fine for now. I agree, though: do you have any suggestion on what to do when we fetch an invalid record?

app/jobs/import_data_from_rapids_job.rb
FerPerales: @littleforest total records comes from the API response. We will need to to a first call outside the while loop to fetch the number of records

app/jobs/import_data_from_rapids_job.rb
littleforest: If we're fine just skipping it, then this is fine and we can leave as-is. Otherwise, do we want to send the data to the error subscriber so we can investigate to see if there are other changes that either need to be made to the API, or will help us realize we aren't getting data from a correct field, for example?

app/jobs/import_data_from_rapids_job.rb
littleforest: Ah I see, my suggestion will definitely not work then.

app/jobs/import_data_from_rapids_job.rb
FerPerales: @littleforest I asked Patrick on Slack 😁

littleforest (approved):
:shipit:

PR #685, opened 2024-03-11
Handle documents for standards without data import or source file

FerPerales:

Occupation Standards created from RAPIDS API do not have any source file or standard import associated. This PR handles this case to make `OccupationStandard#public_document?` and `OccupationStandard#source_file` work

### DIFF
--- a/app/models/occupation_standard.rb
+++ b/app/models/occupation_standard.rb
@@ -254,7 +254,7 @@ class OccupationStandard < ApplicationRecord
   end
 
   def public_document?
-    !!(source_file&.public_document || standards_import&.public_document)
+    (source_file&.public_document || standards_import&.public_document).present?
   end
 
   def original_file_url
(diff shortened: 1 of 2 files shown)


app/models/occupation_standard.rb
```diff
@@ -250,11 +250,11 @@ def related_job_titles
   end
 
   def source_file
-    data_import.source_file
+    data_import&.source_file
   end
 
   def public_document?
-    source_file.public_document || standards_import.public_document
+    !!(source_file&.public_document || standards_import&.public_document)
```
littleforest, line 257: I don't love `!!`. Can we do something more explicit like:

```
doc = source_file&.public_document || standards_import&.public_document
doc.present?
```

FerPerales: @littleforest updated

littleforest (approved):

PR #691, opened 2024-03-14
RAPIDS API - Import Competencies

FerPerales:
[Asana ticket](https://app.asana.com/0/1203289004376659/1206821917760869/f)

WorkProcesses from RAPIDS API have an attributes "tasks", that maps to our Competency model. It is an array with one string separated by semicolons. It may also be an empty string

```
      "dwas": [
        {
          "title": "Safety",
          "minHours": 2000,
          "maxHours": 3000,
          "tasks": [
            "test 1; test 2; test 3"
          ]
        }
      ]
```

### DIFF
--- a/app/models/rapids/work_process.rb
+++ b/app/models/rapids/work_process.rb
@@ -1,11 +1,37 @@
 module RAPIDS
   class WorkProcess
-    def self.initialize_from_response(work_process_data)
-      ::WorkProcess.new(
-        title: work_process_data["title"],
-        minimum_hours: work_process_data["minHours"],
-        maximum_hours: work_process_data["maxHours"]
-      )
+    class << self
+      def initialize_from_response(work_process_data)
+        tasks = work_process_data["tasks"]
+
+        work_process = ::WorkProcess.new(
+          title: work_process_data["title"],
+          minimum_hours: work_process_data["minHours"],
+          maximum_hours: work_process_data["maxHours"]
+        )
+
+        if competencies_available?(tasks)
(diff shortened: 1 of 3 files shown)


app/models/rapids/work_process.rb
```diff
@@ -1,11 +1,37 @@
...
+      end
+
+      private
+
+      def competencies_available?(tasks)
+        tasks&.first&.present?
+      end
+
+      def process_competencies(tasks)
+        tasks_titles = tasks.first.split("; ")
```
littleforest, line 28: Can we split on optional whitespace in case the whitespace is sometimes missing? `split(/\s+;\s+/)`

FerPerales: @littleforest thanks! Added a test for this

spec/models/rapids/work_process_spec.rb
```diff
@@ -16,5 +16,37 @@
       expect(work_process.minimum_hours).to eq 20
       expect(work_process.maximum_hours).to eq 200
     end
+
+    it "returns work process with competencies if available" do
+      work_process_response = create(
+        :rapids_api_detailed_work_activity,
+        tasks: ["Competency #1; Competency #2; Competency #3"]
```
littleforest, line 23: Looks good! Can you remove a space in one of these to check the new regex?
```suggestion
        tasks: ["Competency #1; Competency #2;Competency #3"]
```

littleforest (approved):

PR #692, opened 2024-03-14
Add rake task on production to back extract bulletins

littleforest:
[Asana ticket](https://app.asana.com/0/1203289004376659/1206228142308905/f)

### DIFF
--- a/lib/tasks/deployment/20240314201546_back_extract_bulletin_attachments.rake
+++ b/lib/tasks/deployment/20240314201546_back_extract_bulletin_attachments.rake
@@ -1,0 +1,23 @@
+namespace :after_party do
+  desc "Deployment task: back_extract_bulletin_attachments"
+  task back_extract_bulletin_attachments: :environment do
+    puts "Running deploy task 'back_extract_bulletin_attachments'"
+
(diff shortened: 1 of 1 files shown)


lib/tasks/scraper.rake
```diff
@@ -13,4 +13,21 @@ namespace :scraper do
       puts "Use FORCE=true to force this task"
     end
   end
+
+  desc "one-off task to extract attachments from bulletins"
+  task back_extract_bulletins: :environment do
+    # Checking for bulletin: false since we want to back extract Bulletins
+    # that existed before we added that flag.
+    StandardsImport.where(bulletin: false).where("notes LIKE ?", "%BulletinsJob%").find_each do |standards_import|
```
FerPerales, line 21: @littleforest should we add an `includes` to load any additional association and avoid an n+1 query?

littleforest: In the Administrate PR that I just merged, I added [this includes](https://github.com/ApprenticeshipStandardsDotOrg/ApprenticeshipStandardsDotOrg/blob/main/app/models/standards_import.rb#L23). I'm having trouble getting an `includes` to work directly on the StandardsImport class.

PR #706, opened 2024-03-20
Add bulletin field to source_files table and populate

littleforest:
We would like to know which source file was the original source file downloaded through the `Scraper::ApprenticeshipBulletinsJob`, but currently that is not possible as the `bulletin` field is only on the parent `StandardsImport` model and the standards_import record also holds all of the extracted files. This information will be helpful when we write a task to finish converting all word files to pdf, as we can ensure that we can skip any original bulletin files.

This adds a `bulletin` field to the `source_files` table and adds a one-time rake task to populate the field.


FerPerales (approved):
:shipit:

PR #707, opened 2024-03-21
Convert remaining word files to pdf

littleforest:
[Asana ticket](https://app.asana.com/0/1203289004376659/1206663256915727/f)

This ensures that any assignee on a word source file is transferred to the converted pdf file, and that the assignee is cleared from the now-archived word file.

This also adds a one-time rake task to loop through all StandardsImports and convert any remaining word files to pdf.

### DIFF
--- a/app/jobs/create_source_file_job.rb
+++ b/app/jobs/create_source_file_job.rb
@@ -33,11 +33,15 @@ class CreateSourceFileJob < ApplicationJob
     end
 
     if original_source_file.present?
-      source_file.update!(original_source_file:)
+      source_file.update!(
+        original_source_file:,
+        assignee: original_source_file.assignee
+      )
       original_source_file.update!(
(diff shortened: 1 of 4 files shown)


app/jobs/create_source_file_job.rb
```diff
@@ -33,11 +33,15 @@ def maybe_link_to_original_source_file(source_file, linkable_word_files)
     end
 
     if original_source_file.present?
-      source_file.update!(original_source_file:)
+      source_file.update!(
+        original_source_file:,
+        assignee: original_source_file.assignee
```
FerPerales, line 38: @littleforest is it guaranteed that every source file has an assignee?

littleforest: No, but then it will just assign `nil`, which is fine.

FerPerales (approved):

PR #714, opened 2024-04-01
Add RAPIDS::API#documents

FerPerales:
[Asana ticket](https://app.asana.com/0/1203289004376659/1206821917760866/f)

Add `documents` method to fetch from RAPIDS API. Documents must be consumed using a POST request, as per API docs

This method will be used during RAPIDS import process to attach this document to its occupation standard


littleforest (approved):
Is the documents endpoint actually used to post any data, or is it actually acting as a get request?

FerPerales:
> Is the documents endpoint actually used to post any data, or is it actually acting as a get request?

@littleforest It acts as a GET

PR #716, opened 2024-04-02
Attach documents from RAPIDS API

FerPerales:
[Asana ticket](https://app.asana.com/0/1203289004376659/1206821917760866/f)

RAPIDS API provides access to occupation standard documents by using their docs end-point with a POST request.

This PR add the ability to determine if a standard has a document and, if available, attaches this document as the redacted file for that specific occupation standard.

### DIFF
--- a/spec/jobs/import_data_from_rapids_job_spec.rb
+++ b/spec/jobs/import_data_from_rapids_job_spec.rb
@@ -176,6 +176,113 @@ RSpec.describe ImportDataFromRAPIDSJob, type: :job do
           change(WorkProcess, :count).by(0)
       end
     end
+
+    context "wps document" do
+      context "when document is mark as not uploaded" do
+        it "does not attach a document" do
+          stub_get_token!
+          create_registration_agency_for("MI", agency_type: :oa)
+
+          occupation_standard_response = create_list(
+            :rapids_api_occupation_standard,
+            1,
+            :competency,
+            isWPSUploaded: false
+          )
+          rapids_response = create(:rapids_response, totalCount: 1, wps: occupation_standard_response)
+
+          stub_rapids_api_response(
+            {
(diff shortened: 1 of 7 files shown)


spec/jobs/import_data_from_rapids_job_spec.rb
```diff
@@ -186,17 +176,137 @@
...
+      end
+    end
   end
 end
 
 def stub_get_token!
   allow_any_instance_of(RAPIDS::API).to receive(:get_token!).and_return "xxx"
 end
 
+def create_registration_agency_for(state_abbreviation, agency_type:)
```
mike-burns, line 293: Is it worthwhile to push this into factory_bot, e.g.

```ruby
FactoryBot.define do
  factory :registration_agency do
    traits_for_enum :agency_type, RegistrationAgency.agency_types

    ### remove `state` association, build it manually instead
    transient do
      state_abbreviation { nil }
    end

    agency_type { :oa }

    trait :for_national_program do
      state { nil }
    end

    after(:build) do |registration_agency, context|
      state_attribs = {}

      if context.state_abbreviation
        state_attribs[:abbreviation] = context.state_abbreviation
      end

      registration_agency.state = FactoryBot.build(:state, **state_attribs)
    end
  end
end
```

Then we can do

```ruby
create(:registration_agency, state_abbreviation: "MI", agency_type: :oa)
```

(I haven't tried this code, it might not work.)

mike-burns (approved):
LGTM. I had one thought about moving more things into factory_bot.

spec/jobs/import_data_from_rapids_job_spec.rb
FerPerales: @mike-burns I like that! I'll create another PR with the suggestion. Thanks!

PR #722, opened 2024-04-10
Move registration agency creation to factory

FerPerales:
We usually have to create an state in order to create a registration agency for that state. This PR introduces a transient called `for_state_abbreviation`, which allows to set an specific state abbreviation directly in the factory.

### DIFF
--- a/spec/system/occupation_standards/index_spec.rb
+++ b/spec/system/occupation_standards/index_spec.rb
@@ -564,7 +564,6 @@ RSpec.describe "occupation_standards/index" do
 
     it "filters standards with state shortcode" do
       stub_feature_flag(:use_elasticsearch_for_search, true)
-      # washington = create(:state, name: "Washington", abbreviation: "WA")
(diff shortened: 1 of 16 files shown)


spec/system/occupation_standards/index_spec.rb
```diff
@@ -565,8 +564,8 @@
 
     it "filters standards with state shortcode" do
       stub_feature_flag(:use_elasticsearch_for_search, true)
-      washington = create(:state, name: "Washington", abbreviation: "WA")
-      washington_registration_agency = create(:registration_agency, state: washington)
+      # washington = create(:state, name: "Washington", abbreviation: "WA")
```
littleforest, line 567: Remove comment

littleforest (approved):
Nice!

FerPerales:
> This came out quite nicely.

@mike-burns thanks for the suggestion and thanks to Neil for showing me that you can use transients as part of other attributes, as in this case 😄

PR #723, opened 2024-04-11
Introduce import tree

mike-burns:
Document imports come in via two ways:

- User uploads a document directly, typically a PDF or Docx
- We fetch data from an endpoint, typically PDF or Docx

The Docx that we fetch sometimes has attachments, which we then continue to process. It can look like this:

```
Docx (listing)
    |
   / \
Docx  Docx
  |    |
 PDF  PDF
```

A `StandardsImport` model can contain many of these trees, if we process multiple API endpoints or if the user uploads many documents.

An additional thing to note is that knowing whether something is a Docx listing, Docx, Doc, or PDF is itself a processing step. At first it's an uncategorized attachment.

Therefore, we can imagine this structure:

```
                  StandardsImport
                         |
     /-------------------+-------------------\
     |                   |                   |
Uncategorized      Uncategorized       Uncategorized
    |                    |                   |
Docx (listing)          PDF                 Doc
    |                                        |
   / \                                      PDF
Docx  Docx
  |    |
 PDF  PDF
```

And so we get our processing rules:

- An Uncategorized processes into a Docx, Docx listing, Doc, or PDF
- A Docx listing processes into many imports, which are Docx, Doc, or PDF
- A Docx and Doc process into a PDF

And our assumptions:

- A StandardsImport has many imports, all of which are Uncategorized
- A PDF terminates

---

This commit is the first step in implementing this. It add the tree, but does not yet connect it. We have documented connection tasks in Asana.

The tree is single table inheritance (STI). This allows the StandardsImport#imports to be of any type, and each import can have any parent.

STI brings some constraints:

- The factory_bot definitions must be separate for each child, and must define a `type` column.
- The `belongs_to :parent` must be in the parent (`Import`). To quote [the Rails docs]:
    > Using polymorphic associations in combination with single table
    > inheritance (STI) is a little tricky. In order for the
    > associations to work as expected, ensure that you store the base
    > model for the STI models in the type column of the polymorphic
    > association. To continue with the asset example above, suppose
    > there are guest posts and member posts that use the posts table
    > for STI. In this case, there must be a type column in the posts
    > table.

[the Rails docs]: https://api.rubyonrails.org/classes/ActiveRecord/Associations/ClassMethods.html#module-ActiveRecord::Associations::ClassMethods-label-Polymorphic+Associations

STI also means that the one table must cater to every child:

- `assignee`, `public_document`, `courtesy_notification`, `metadata` - needed for PDF
- `status`, `parent`, `type` - needed for all children

---

[Asana ticket](https://app.asana.com/0/1203289004376659/1207002310256042)

Co-authored-by: Fernando Perales <fer@thoughtbot.com>
Co-authored-by: Jeanine Soterwood <jeanine@thoughtbot.com>

### DIFF
--- a/app/models/imports/doc.rb
+++ b/app/models/imports/doc.rb
@@ -1,0 +1,6 @@
+module Imports
+  class Doc < Import
+    has_one :pdf, as: :parent, dependent: :destroy, autosave: true
+    has_one_attached :file
+  end
+end
--- a/app/models/standards_import.rb
+++ b/app/models/standards_import.rb
@@ -1,5 +1,6 @@
 class StandardsImport < ApplicationRecord
   has_many_attached :files
+  has_many :imports, as: :parent, dependent: :destroy
 
   enum courtesy_notification: [:not_required, :pending, :completed], _prefix: true
 
--- a/app/models/import.rb
+++ b/app/models/import.rb
@@ -1,0 +1,3 @@
+class Import < ApplicationRecord
+  belongs_to :parent, polymorphic: true
+end
--- a/app/models/imports/docx.rb
+++ b/app/models/imports/docx.rb
@@ -1,0 +1,6 @@
+module Imports
+  class Docx < Import
+    has_one :pdf, as: :parent, dependent: :destroy, autosave: true
+    has_one_attached :file
+  end
+end
--- a/app/models/imports/docx_listing.rb
+++ b/app/models/imports/docx_listing.rb
@@ -1,0 +1,6 @@
+module Imports
+  class DocxListing < Import
+    has_many :imports, as: :parent, dependent: :destroy, autosave: true
+    has_one_attached :file
+  end
+end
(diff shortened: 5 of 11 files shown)


app/models/imports/doc.rb
```diff
@@ -0,0 +1,6 @@
+module Imports
+  class Doc < Import
+    has_one :pdf, as: :parent, dependent: :destroy, autosave: true
```
FerPerales, line 3: @mike-burns it is the autosave the reason it was not working last time?

FerPerales (approved):

app/models/standards_import.rb
```diff
@@ -1,5 +1,6 @@
 class StandardsImport < ApplicationRecord
   has_many_attached :files
+  has_many :imports, dependent: :destroy, as: :parent
```
littleforest, line 3: Super minor and picky, but can we put the `as: :parent` after the `has_many :imports` to have consistency with the other files?

littleforest (approved):
Looks great!

app/models/imports/doc.rb
mike-burns: The autosave, and also the `belongs_to :parent` needed to be in the Import class.

PR #726, opened 2024-04-11
Handle duplicates from RAPIDS API

FerPerales:
[Asana ticket](https://app.asana.com/0/1203289004376659/1206983146425385/f)

While importing data from RAPIDS API, they may be duplicates. This PR checks if an OccupationStandard is already present by searching for a record with the same title and organization name


spec/jobs/import_data_from_rapids_job_spec.rb
```diff
@@ -98,6 +98,35 @@
       end
     end
 
+    context "when receiving an existing standard" do
+      it "does not create a duplicated record" do
+        stub_get_token!
+        create(:registration_agency, for_state_abbreviation: "MI", agency_type: :oa)
+
+        occupation_standard_response = create_list(
```
littleforest, line 106:
```suggestion
        occupation_standard_response = create_pair(
```
and can remove the "2" on line 108.

littleforest (approved):

PR #728, opened 2024-04-12
Imported file process

mike-burns:
Imports::Docx#process

We process Docx files identically to Doc files.

<https://app.asana.com/0/1203289004376659/1207002310256047>

---

Imports::Doc#process

Process a Doc import file by converting it to PDF.

We want to track whether the processing has finished, so we add new
columns `processed_at` and `processing_errors`. Only one is present at a
time.

We can also start to make use of the `status` column, to flag import
files that need attention.

Much of the work for this is already done in `DocToPdfConverter`.
However, that service hardcodes the idea of a `SourceFile`. Instead, we
want a service that is divorced from the database entirely: it should
take a string representing some sort of unique ID, and the file to
process. On success we need the filename of the PDF output.

So to ease the transition, make a `ConvertDocToPdf` service and use it
from `DocToPdfConverter`. Many of the tests are identical, just simpler.

While here, remove the unused `DocToPdfConverter.convert_all` method,
which attempts to enqueue a job but that job misspelled `#perform`.

<https://app.asana.com/0/1203289004376659/1207002310256046>

---

Imports::Pdf#process

The goal of the `#process` method is to get an imported file one step
closer to a PDF. Therefore, processing a PDF does nothing.

All imported files must be processable. Extract a shared example to
convey this shared interface.

<https://app.asana.com/0/1203289004376659/1207002310256044>

---

Imports::Uncategorized#process

To process an uncategorized import, we have three steps:

1. Categorize it,
2. Process it according to the category,
3. Mark ourselves as processed.

To categorize an import, we need to know the filename suffix and whether
it's part of a listing (`Scraper::ApprenticeshipBulletinsJob`):

- `.doc` and `.pdf` are straightforward.
- `.docx` is either an Imports::Docx or an Imports::DocxListing,
  depending on whether we came from ApprenticeshipBulletinsJob.

So extend `#process` to take flags that we can pass through as
appropriate. Each import can make use of the flag as it sees fit. This
allows for extension later as we add more ways to acquire imports.

<https://app.asana.com/0/1203289004376659/1207002310256043>

---

DocxListing#process

To process a DocxListing (a Docx file with attachments), first
split the attachments out. For each attachment, create an
`Imports::Uncategorized` child, and then process all children.

Along the way we realized that we don't handle the case where the
Uncategorized file is of an unknown type. For that, we raise.

The DocxListing#process method does not handle the error; any invalid
attachment stops everything. If we ever start processing files with
unexpected types we can reconsider that (though honestly we'll probably
want to process those files instead).

<https://app.asana.com/0/1203289004376659/1207002310256045>


app/models/imports/pdf.rb
```diff
@@ -1,5 +1,8 @@
 module Imports
   class Pdf < Import
     has_one_attached :file
+
+    def process
```
mike-burns, line 5: This should probably set `procesed_at` and `status`.

app/models/imports/docx.rb
```diff
@@ -2,5 +2,28 @@ module Imports
...
+          processed_at: Time.now,
+          processing_errors: nil,
+          status: :completed,
+        )
+      end
+    rescue ConvertDocToPdf::PdfConversionError => e
+      update!(
+        processed_at: nil,
+        processing_errors: e.message,
+        status: :needs_support,
```
mike-burns, line 25: Is this `status` value correct? Are any of these `status` values correct?

(They are meant to mimic the SourceFile status value.)

app/services/doc_to_pdf_converter.rb
```diff
@@ -1,13 +1,6 @@
 class DocToPdfConverter
```
mike-burns, line 1: The goal is to delete this file.

app/models/imports/docx.rb
FerPerales: @mike-burns Right now, I can't think of a reason to change them, but I'm sure once we go deeper, there may be some statuses that won't be valid for some imports. 

For this specific case, needs support seems correct. @littleforest can you help us with this one?

FerPerales (approved):
Looks good. There are some broken specs and linter errors and we still need to confirm the `status`, but this PR is a good start

app/models/imports/doc.rb
```diff
@@ -2,5 +2,28 @@ module Imports
...
+      output_pdf_path = ConvertDocToPdf.call(id, file)
+
+      transaction do
+        create_pdf!(
+          file: File.open(output_pdf_path),
+          public_document: public_document,
+          metadata: metadata,
+        )
+        update!(
+          processed_at: Time.now,
```
littleforest, line 16:
```suggestion
          processed_at: Time.current,
```

app/models/imports/docx.rb
```diff
@@ -2,5 +2,28 @@ module Imports
...
+      output_pdf_path = ConvertDocToPdf.call(id, file)
+
+      transaction do
+        create_pdf!(
+          file: File.open(output_pdf_path),
+          public_document: public_document,
+          metadata: metadata,
+        )
+        update!(
+          processed_at: Time.now,
```
littleforest, line 16:
```suggestion
          processed_at: Time.current,
```

app/models/imports/docx.rb
littleforest: The `needs_support` status is really to indicate that help is needed deciphering how data should get transferred from the doc/pdf file into the spreadsheet that is uploaded. So maybe we need a new status, something like `backend_support`, or `convert_failure`?

app/services/doc_to_pdf_converter.rb
```diff
@@ -1,13 +1,6 @@
...
   TMP_DIRECTORY = "./tmp/soffice".freeze
 
-  def self.convert_all
```
littleforest, line 4: There is a `lib/tasks/redacted_files.rake` file that was being used as a one-off task to call this that can be cleaned up.

mike-burns:
TODO `#process` needs to recur down the tree.

app/models/imports/doc.rb
```diff
@@ -2,5 +2,28 @@ module Imports
...
+          processing_errors: nil,
+          status: :completed,
+        )
+      end
+    rescue ConvertDocToPdf::PdfConversionError => e
```
mike-burns, line 21: We should probably catch all exceptions, log them, and then re-raise them.

mike-burns:
TODO: `app/jobs/scraper/apprenticeship_bulletins_job.rb` adds an Excel file -- should that be an Imports::Excel ?

mike-burns:
@littleforest ready for final review, please.

littleforest (approved):
🎉  Can't wait to start implementing this!

PR #735, opened 2024-04-18
Remove show_recently_added_section flag

jenlu33:
[Asana ticket](https://app.asana.com/0/1203289004376659/1206186965082996/f)


spec/system/pages/home_spec.rb
```diff
@@ -111,7 +111,6 @@
   context "with Elasticsearch enabled", :elasticsearch do
     before(:each) do |example|
       stub_feature_flag(:use_elasticsearch_for_search, true)
-      stub_feature_flag(:show_recently_added_section, true)
```
FerPerales, line 114: @jenlu33 there is another test in this file that uses the flag in line 282. Can we also remove that one?

littleforest (approved):

PR #737, opened 2024-04-23
Store sponsor number from rapids api

FerPerales:
[Asana ticket](https://app.asana.com/0/1203289004376659/1207105403377729/f)

This PR adds a JSONB column called metadata to OccupationStandard. This is used to store the complete response from RAPIDS API.

This also adds a method to access the sponsor number from the response mentioned above

Why do we want to store the complete response instead of the sponsor number?

By having the data, we can now do some queries to extract some requested statistics

### DIFF
--- a/app/models/occupation_standard.rb
+++ b/app/models/occupation_standard.rb
@@ -404,10 +404,6 @@ class OccupationStandard < ApplicationRecord
     base_onet
   end
 
-  def sponsor_number
-    metadata["sponsorNumber"]
-  end
-
   private
 
   def national?
--- a/app/models/occupation_standard.rb
+++ b/app/models/occupation_standard.rb
@@ -404,6 +404,10 @@ class OccupationStandard < ApplicationRecord
     base_onet
   end
 
+  def sponsor_number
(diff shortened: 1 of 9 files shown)


mike-burns:
> By having the data, we can now do some queries to extract some requested statistics

Do we want to store that data, too?

FerPerales:
> > By having the data, we can now do some queries to extract some requested statistics
> 
> Do we want to store that data on the table, too?

Yes. We are doing that here https://github.com/ApprenticeshipStandardsDotOrg/ApprenticeshipStandardsDotOrg/pull/737/files#diff-d1d1b9ab2aee01570c0fb7b357d95cf9949b33024cf9d75468f7e69f8a68aa5cR27

mike-burns:
I meant, do we want to add columns for those other fields?

FerPerales:
> I meant, do we want to add columns for those other fields?

Probably not. We want to have the payload to check that we imported the payload correctly. For this ticket, we only want to have "access" to the sponsor number to verify that it matches the expected registration state

littleforest (approved):

PR #738, opened 2024-04-23
Handle retry when token expires

FerPerales:
It may happen that the OAuth token expires during the importing process. This PR handles that case and refreshes the token prior to retry

### DIFF
--- a/app/jobs/import_data_from_rapids_job.rb
+++ b/app/jobs/import_data_from_rapids_job.rb
@@ -28,7 +28,7 @@ class ImportDataFromRAPIDSJob < ApplicationJob
       if tries < MAXIMUM_TRIES
         @service.get_token!
         tries += 1
-        Rails.logger.error "OAuth2::Error. Retrying #{tries} / #{MAXIMUM_TRIES}"
+        Rails.logger.warn "[RAPIDSAPI] OAuth2::Error. Retrying #{tries} / #{MAXIMUM_TRIES}"
         retry
       end
     end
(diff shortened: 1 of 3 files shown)


app/jobs/import_data_from_rapids_job.rb
```diff
@@ -4,22 +4,33 @@ class ImportDataFromRAPIDSJob < ApplicationJob
...
+        start_index += PER_PAGE_SIZE
 
-      break if start_index > total_records
+        break if start_index > total_records
+      end
+    rescue OAuth2::Error
+      if tries < MAXIMUM_TRIES
+        @service.get_token!
+        tries += 1
+        Rails.logger.error "OAuth2::Error. Retrying #{tries} / #{MAXIMUM_TRIES}"
```
mike-burns, line 31: Real pedantic here: is `Rails.logger.error` right, or is `warning` better? I know you're handling an error, but it feels like `error` is right for when we run out of retries, `warn` is right for when we are retrying.

What do you think?

app/jobs/import_data_from_rapids_job.rb
```diff
@@ -4,22 +4,33 @@ class ImportDataFromRAPIDSJob < ApplicationJob
...
+        retry
+      end
```
mike-burns, line 33: Do we want an `else`?

app/jobs/import_data_from_rapids_job.rb
FerPerales: @mike-burns that makes sense!

littleforest (approved):

PR #751, opened 2024-05-01
Add DocxListingSplitter, extract OLE storage

mike-burns:
Extract a `DocxListingSplitter` from `ExportFileAttachmentsJob`. This service is only in charge of turning an ActiveStorage file into its attachments, and knows nothing of the database. We need this for a future refactoring.

While working on this, we learned that the `.bin` files attached to Docx docs are _not_ PDFs. They are OLE storage files
(`application/x-ole-storage`) which _contain_ PDFs. Since anything that works with PDFs expects a PDF, passing the OLE storage file around has variable results:

- macOS's Preview program can open OLE storage files containing PDFs, and treats them as if they are PDFs.
- MATE's Atril program does not know what to do with them.
- soffice is unable to convert them to PDF.
- Marcel, the MIME detector in Rails, reports that they are `application/x-ole-storage`.
- most crucially, `SourceFile.pdf_attachment` does not recognize them as PDFs (since it relies on Marcel), meaning these PDF attachments are effectively dropped.

The solution is to assume all `.bin` files are OLE storage files wrapping PDFs. We can use the ruby-ole library to extract the `CONTENTS` key and write that to a `.pdf` file, then the rest of the system can ignore the `.bin`.

We assume all `.bin` files are PDFs because so far this has been the case. I tried to find an OLE storage mechanism for knowing the MIME type of the attachments but didn't find it.


app/services/docx_listing_splitter.rb
```diff
@@ -0,0 +1,86 @@
...
+    file_names = unzip_attachments_and_list_file_names(zip_file)
+
+    block.call(file_names)
+  ensure
+    delete_extracted_files(zip_file)
+  end
+
+  private
+
+  def create_zip_version_of_source_file
```
littleforest, line 24: Should this method name be renamed (and the `zip_file` variable name), since we're not creating a zip version anymore? Though we are still using the `Zip::File` to unzip it, I feel like it is a little confusing since there is no actual Zip file creation happening.

littleforest (approved):
So glad you ran into the issues with the `.bin` files that we overlooked on our Macs. Looks great! Just one minor naming question/suggestion.

app/services/docx_listing_splitter.rb
mike-burns: Right, interesting point. Some thoughts:

- we are definitely writing out a Zip file, but its filename ends in `.docx`, so calling the variable `zip_file` could make sense.
- we aren't creating a Zip version of a SourceFile, so the method name doesn't make sense.

I'm going to rename the method to `download_file`.

littleforest: Sounds good. I think leaving the file name as `zip_file` works and won't be confusing as we are unzipping it in the next method.

PR #757, opened 2024-05-02
Modify Bulletins scraper job to use the new Imports model

littleforest:
This creates a new service to process one row of the scraped Apprenticeship Standards Bulletin CSV file. Each row creates one `StandardsImport` record and a child `Imports::Uncategorized record`. The `process` method is then triggered on the Imports::Uncategorized record to start the process of getting to a `Imports::Pdf` record.

[Asana ticket](https://app.asana.com/0/0/1207002310256048/f)


app/services/process_apprenticeship_bulletin.rb
```diff
@@ -0,0 +1,43 @@
...
+      bulletin: true,
+      metadata: {date: date}
+    )
+
+    if standards_import.new_record?
+      standards_import.save!
+
+      filename = File.basename(URI.decode_uri_component(uri))
+      if standards_import.files.attach(io: URI.parse(uri).open, filename: filename)
+        import = standards_import.imports.create(
```
mike-burns, line 29: Since the goal is to remove `standards_import.files`, what do you think of centering this around the `standards_import.imports` instead? Something like:


```ruby
import = standards_import.imports.create!(
  type: "Imports::Uncategorized",
  file: URI.parse(uri).open,
  public_document: true,
  metadata: standards_import.metadata
)
import.process(listing: true)

standards_import.files.attach(standards_import.imports.first.file)
```

This way we can easily delete that last line later.

app/services/process_apprenticeship_bulletin.rb
```diff
@@ -0,0 +1,43 @@
...
+          type: "Imports::Uncategorized",
+          file: standards_import.files.first.blob,
+          public_document: true,
+          metadata: standards_import.metadata
+        )
+        import.process
```
mike-burns, line 35: I think these are Docx listings, right? If so, we should pass `listing: true` here.

mike-burns (commented):
That was fast!

mike-burns (approved):
Nice!

PR #761, opened 2024-05-06
Use tree structure for files uploaded through the UI

littleforest:
For files uploaded through the UI, we now create
`Imports::Uncategorized` records instead of SourceFile
 records. These Uncategorized records will not
immediately be processed. The processing will
be handled by a rake task in a future PR.

[Asana ticket](https://app.asana.com/0/0/1207002310256049/f)

### DIFF
--- a/app/controllers/standards_imports_controller.rb
+++ b/app/controllers/standards_imports_controller.rb
@@ -10,9 +10,6 @@ class StandardsImportsController < ApplicationController
 
     unless user_signed_in?
       @standards_import.courtesy_notification = :pending
-      @standards_import.imports.each do |import|
-        import.courtesy_notification = :pending
-      end
     end
 
     if @standards_import.save
--- a/app/controllers/standards_imports_controller.rb
+++ b/app/controllers/standards_imports_controller.rb
@@ -10,6 +10,9 @@ class StandardsImportsController < ApplicationController
 
     unless user_signed_in?
       @standards_import.courtesy_notification = :pending
+      @standards_import.imports.each do |import|
+        import.courtesy_notification = :pending
+      end
     end
 
     if @standards_import.save
--- a/app/models/standards_import.rb
+++ b/app/models/standards_import.rb
@@ -48,6 +48,17 @@ class StandardsImport < ApplicationRecord
     source_files.count == source_files.count { |source_file| source_file.courtesy_notification_completed? }
   end
 
+  def files=(files)
+    files.each do |file|
+      imports.build(
+        type: "Imports::Uncategorized",
+        status: :pending,
+        public_document: public_document,
+        file: file
+      )
+    end
+  end
+
   def file_count
     files.count
   end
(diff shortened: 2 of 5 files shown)


app/controllers/standards_imports_controller.rb
```diff
@@ -10,6 +10,9 @@ def create
 
     unless user_signed_in?
       @standards_import.courtesy_notification = :pending
+      @standards_import.imports.each do |import|
+        import.courtesy_notification = :pending
```
mike-burns, line 14: I'd like to learn more about this step. Is this something we can set from `StandardsImport#files=`?

mike-burns (approved):
You're making this look easy!

app/controllers/standards_imports_controller.rb
littleforest: I tried that first, but since we are not passing in `courtesy_notification` as a form field, we can't set it in `files=`, which gets called on line 9. Any other ideas? We could pass it in as a form field, but that bugs me a bit since it isn't necessary to come in on the form.

app/models/standards_import.rb
```diff
@@ -44,6 +44,18 @@ def has_notified_uploader_of_all_conversions?
...
 
+  def files=(files)
+    files.each do |file|
+      imports.build(
+        type: "Imports::Uncategorized",
+        status: :pending,
+        public_document: public_document,
+        file: file
+      )
+    end
```
littleforest, line 55: This is messing up a bunch of specs, so I will probably wait to merge this as one of the last pieces to fix.

app/controllers/standards_imports_controller.rb
mike-burns: I came up with something uglier but deleted that.

Is this something we can set before_save? Spitballing here:

```ruby
class Imports < ApplicationRecord
  before_save :set_courtesy_notification

  def standard_courtesy_notification
    parent.standard_courtesy_notification
  end

  private

  def set_courtesy_notification
    self.courtesy_notification = standard_courtesy_notification
  end
end
```

```ruby
  def standard_courtesy_notification
    courtesy_notification
  end
end
```

Then we can keep line 12 in the controller but drop 13-15.

Is that worse?

littleforest: Oh I like that way better. Thanks!

littleforest: Fixed in c86aa55c70d6fc48338b52ce107f9e1b856b9286

mike-burns (approved):
LGTM, and I looked at the test failures and agree: this might Just work later.

mike-burns:
I think this needs to go in before the backfill, which is the next item in the todo list. I'll spend a few minutes looking at the test failures to see if it's actually something we can quickly fix.

littleforest:
> I think this needs to go in before the backfill

What piece here is needed for the backfill? I actually think the tickets around linking up the DataImport records and allowing converters to claim (assign themselves) will need to happen soon so we can start transitioning the converters over to the new system.

mike-burns:
This PR augments the file upload to also store into the new imports tree. The backfill's goal is to add import trees for all existing StandardsImports. If we merge the backfill before we merge this, then we'll miss out on files uploaded between the backfill and this merge.

PR #767, opened 2024-05-07
Add Imports to Administrate dashboard

littleforest:
This adds the index and show pages for Imports. It is currently behind a `:show_imports_in_administrate` feature flag so as not to confuse the converters. Admin users will see all Imports, but Converters will only see the final `Import::Pdf` type.

Screenshot 2024-05-07 at 3 54 29 PM

Screenshot 2024-05-08 at 8 57 11 AM

[Asana ticket](https://app.asana.com/0/0/1207225092876016/f)

### DIFF
--- a/app/dashboards/import_dashboard.rb
+++ b/app/dashboards/import_dashboard.rb
@@ -1,0 +1,91 @@
+require "administrate/base_dashboard"
+
+class ImportDashboard < Administrate::BaseDashboard
+  # ATTRIBUTE_TYPES
+  # a hash that describes the type of each of the model's fields.
+  #
+  # Each different type represents an Administrate::Field object,
+  # which determines how the attribute is displayed
+  # on pages throughout the dashboard.
+  ATTRIBUTE_TYPES = {
+    id: Field::String,
+    assignee: Field::BelongsTo,
+    type: Field::String,
+    courtesy_notification: Field::Select.with_options(searchable: false, collection: ->(field) { field.resource.class.send(field.attribute.to_s.pluralize).keys }),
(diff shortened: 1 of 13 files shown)


app/views/admin/imports/show.html.erb
```diff
@@ -0,0 +1,77 @@
...
+          <dt class="attribute-label" id="<%= attribute.name %>">
+            <%= t(
+                  "helpers.label.#{resource_name}.#{attribute.name}",
+                  default: page.resource.class.human_attribute_name(attribute.name)
+                ) %>
+          </dt>
+
+          <dd class="attribute-data attribute-data--<%= attribute.html_class %>">
+            <% if attribute.name == "parent" %>
+              <% if attribute.data.is_a?(StandardsImport) %>
```
mike-burns, line 65: :+1:

app/dashboards/import_dashboard.rb
```diff
@@ -0,0 +1,90 @@
...
+    file
+    assignee
+    courtesy_notification
+    metadata
+    parent
+    processed_at
+    processing_errors
+    public_document
+    status
+    type
```
mike-burns, line 52: I think this should be more prominent, perhaps after the `id`. What do you think?

mike-burns (commented):
LGTM, had one idea on moving an attribute around.

mike-burns (approved):

app/dashboards/import_dashboard.rb
littleforest: fixed in 1101286e8773b16f52c65e49faeb401917a10cbd

PR #768, opened 2024-05-08
Allow attaching DataImport records to Imports::Pdf records

littleforest:
This allows `DataImport` records to get attached to `Imports::Pdf` records as well as `SourceFile` records through the `:show_imports_in_administrate` feature flag. 

Screenshot 2024-05-08 at 2 19 26 PM

Screenshot 2024-05-08 at 2 20 08 PM

After `DataImport` has been uploaded, here is the `Imports::Pdf` record:

Screenshot 2024-05-08 at 2 20 56 PM

And this is the `DataImport` show page:

Screenshot 2024-05-08 at 2 21 42 PM

[Asana ticket](https://app.asana.com/0/0/1207249791981946/f)


app/models/import.rb
```diff
@@ -1,6 +1,7 @@
 class Import < ApplicationRecord
   belongs_to :parent, polymorphic: true
   belongs_to :assignee, class_name: "User", optional: true
+  has_many :data_imports, inverse_of: "import"
```
mike-burns, line 4: Is it possible (and reasonable) to put this on `Imports::Pdf` instead?

littleforest: I had it on `Imports::Pdf` but it broke Administrate's `import#show` page for non-Imports::Pdf, so I ended up moving it back here. I tried adding 

```
def data_imports
  DataImport.none
end
```

but that didn't work either since it wasn't an association.

mike-burns (approved):
Dang. I think we need to improve Administrate's STI support in the future.

app/models/import.rb
littleforest: @mike-burns what do you think about 7b22826ba15b7f09b0fb189da2499b1b57d27f91 ?

mike-burns: Oooh, cool idea!

I like it a lot. I don't say this often, but would it be helped by a comment above the `-> {none}` explaining what's going on?

littleforest: Added in f1621569b74899a5cad10fe8ff5adb11baa5c00d

PR #770, opened 2024-05-09
Add unfurled status to Import model

littleforest:
This adds an `unfurled` status to be used on newly created `Imports::Uncategorized` records and all children except `Imports::Pdf` records. `Imports::Pdf` will go back to being marked as `pending` upon creation, and will not be marked as `completed` until the corresponding `DataImport` records are processed.

### DIFF
--- a/app/models/imports/uncategorized.rb
+++ b/app/models/imports/uncategorized.rb
@@ -27,7 +27,7 @@ module Imports
 
     def create_child!(**kwargs)
       create_import!(
-        status: :pending,
+        status: child_status,
         assignee_id: assignee_id,
         public_document: public_document,
         courtesy_notification: courtesy_notification,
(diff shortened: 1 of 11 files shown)


mike-burns (approved):

app/models/imports/uncategorized.rb
```diff
@@ -27,7 +27,7 @@ def process(**)
 
     def create_child!(**kwargs)
       create_import!(
-        status: :pending,
+        status: child_status,
```
FerPerales, line 30: @littleforest I don't see any other reference to `child_status`: What is this used for? Will it change?

FerPerales: nvm just found it: i thought it was :child_status instead of a method call 😁

FerPerales (approved):

PR #775, opened 2024-05-14
Establish redacted_pdf for Imports::Pdf model

littleforest:
This adds a `redacted_pdf` file attachment
to the `Imports::Pdf` model. Existing methods
from the `SourceFile` model are copied over
to the `Imports::Pdf` model and modified. This
is the first of multiple PRs to transfer the
redaction feature from `SourceFile` records
to `Imports::Pdf` records.

[Asana ticket](https://app.asana.com/0/1203289004376659/1207319519426700/f)

### DIFF
--- a/app/models/import.rb
+++ b/app/models/import.rb
@@ -7,8 +7,6 @@ class Import < ApplicationRecord
   # of Administrate.
   has_many :data_imports, -> { none }, inverse_of: "import"
 
-  has_one_attached :redacted_pdf
-
   enum :status, [
     :pending,
     :completed,
--- a/app/models/import.rb
+++ b/app/models/import.rb
@@ -7,6 +7,8 @@ class Import < ApplicationRecord
   # of Administrate.
   has_many :data_imports, -> { none }, inverse_of: "import"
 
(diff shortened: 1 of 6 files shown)


app/models/import.rb
```diff
@@ -7,6 +7,8 @@ class Import < ApplicationRecord
   # of Administrate.
   has_many :data_imports, -> { none }, inverse_of: "import"
 
+  has_one_attached :redacted_pdf
```
mike-burns, line 10: Is this on the `Import` instead of `Imports::Pdf` because of Administrate?

mike-burns (commented):
LGTM, just one question about `has_one_attached`.

app/models/import.rb
littleforest: Oops, I thought I put that on `Imports::Pdf`. Thanks for catching. I haven't fit this into Administrate yet, so I might have to move it back, but I would prefer it just in the Pdf model for now.

littleforest: fixed in 1543809da677e571b89847007074233a296b824e

mike-burns (approved):

PR #776, opened 2024-05-15
Allow redacting of private Imports::Pdf files

littleforest:
This adds the ability to redact `Imports::Pdf` documents. This also adds in the filters for displaying redacted vs non-redacted files.

**Admin Index view**
Screenshot 2024-05-15 at 11 55 16 AM

**Admin show view**
Screenshot 2024-05-15 at 11 55 48 AM

**Converter show view**
Screenshot 2024-05-15 at 11 56 01 AM

[Asana ticket](https://app.asana.com/0/1203289004376659/1207319519426700/f)

### DIFF
--- a/app/controllers/admin/redact_files_controller.rb
+++ b/app/controllers/admin/redact_files_controller.rb
@@ -1,8 +1,13 @@
 module Admin
   class RedactFilesController < Admin::ApplicationController
     def new
-      @source_file = SourceFile.find(params[:source_file_id])
+      @record = if Flipper.enabled?(:show_imports_in_administrate)
+        Imports::Pdf.find(params[:import_id])
+      else
+        SourceFile.find(params[:source_file_id])
+      end
       authorize :redact_file, :new?
+
       # Use a custom layout
       render layout: "admin/pdf_editor"
     end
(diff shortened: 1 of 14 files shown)


app/controllers/admin/redact_files_controller.rb
```diff
@@ -1,30 +1,45 @@
...
+            @record.update(redacted_at: Time.current)
+
             render json: {
               message: "Redacted document saved for all occupation standards associated to this source file",
               status: :ok
             }.to_json
           else
             render json: {
-              error: "Redacted document saved for all occupation standards assoaciated to this source file",
+              error: "Redacted document saved for all occupation standards associated to this source file",
```
littleforest, line 42: @FerPerales I just copied this over from SourceFile, but is this error statement correct? Seems like it should say that no redacted file was able to be attached?

FerPerales: @littleforest oops! You are right. The original error message is incorrect

littleforest: fixed in e657003a585bd117f286e67aa6519f5029c4a0ba

mike-burns (approved):

PR #785, opened 2024-05-16
Link occupation standards to imports when feature flag on

littleforest:
When the imports feature flag is on, then an `occupation_standard.source_file` will link to an Import record. While in the short term, this terminology may be confusing, in the long term after we have gotten rid of the SourceFile model, then an `occupation_standard.source_file` will be more apt of a name than `occupation_standard.import`.

This PR is best viewed when hiding whitespace:

Screenshot 2024-05-16 at 1 32 21 PM

[Asana ticket](https://app.asana.com/0/1203289004376659/1207337122136912/f)

### DIFF
--- a/app/models/import.rb
+++ b/app/models/import.rb
@@ -32,11 +32,7 @@ class Import < ApplicationRecord
   def redacted_pdf_url
   end
 
-  def root
-    if parent.is_a?(StandardsImport)
-      parent
-    else
-      parent.root
-    end
+  def import_root
+    parent.import_root
   end
 end
--- a/app/models/import.rb
+++ b/app/models/import.rb
@@ -31,4 +31,12 @@ class Import < ApplicationRecord
 
   def redacted_pdf_url
   end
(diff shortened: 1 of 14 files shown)


app/models/import.rb
```diff
@@ -31,4 +31,12 @@ def redacted_pdf
 
   def redacted_pdf_url
   end
+
+  def root
+    if parent.is_a?(StandardsImport)
```
mike-burns, line 36: Alternative implementation in case you want to avoid an `is_a?` check:

```ruby
class Import < ApplicationRecord
  def import_root
    parent.import_root
  end
end

  def import_root
    self
  end
end
```

mike-burns (approved):

app/models/import.rb
littleforest: Oh yes, that's nicer, thanks! 605c7d581722e48a930bd4c239b397b2212e5e05

PR #790, opened 2024-05-20
Add rake tasks to convert existing files to Import records

littleforest:
This adds a temporary new `source_file` reference field on the `Import` model so we keep track of which source files have corresponding import records. Once we cut over to the new system, this field can be removed.

Three new rake tasks are created. The first to be run will be `source_file:create_imports`. This takes all unarchived SourceFile records that do not have a corresponding import record and creates a new `Imports::Uncategorized` record that links to a SourceFiles attachment. The task leaves the status of the newly created import records as `unfurled` and does not process the imports at this stage.

Once the new `Imports::Uncategorized` records have been created, we run the `import:process_unfurled` rake task. In order to make sure we avoid any duplicate processing caused by race conditions, this task will start with all unfurled `Doc` and `Docx` imports and process them first, which will result in new `Imports::Pdf` records. After this, `DocxListing` records are unfurled, and as part of the `process` method, any new `Doc` and `Docx` records will also be processed. Finally, we process any unfurled `Imports::Uncategorized` records down the tree.

After all the imports have been unfurled, we run `data_import:link_to_import` which will take the linked source_file and find the pdf leaf from the source_file's linked import. It will then set the `import_id` field of the `data_import` record. Once we flip the Imports flag, then the occupation_standards will be automatically linked to the new import records through the `data_import` record.

[Asana ticket](https://app.asana.com/0/1203289004376659/1207249791981940/f)

### DIFF
--- a/lib/tasks/import.rake
+++ b/lib/tasks/import.rake
@@ -8,8 +8,11 @@ namespace :import do
     ].each do |model|
       puts "Processing unfurled #{model} records"
       model.unfurled.find_each.with_index do |import, index|
-        import.process
-        rest(index)
+        begin
+          import.process
+          rest(index)
+        rescue
+        end
       end
 
       sleep 60
@@ -17,9 +20,9 @@ namespace :import do
   end
 
   def rest(index)
-    if (index+1) % 200 == 0
+    if (index+1) % 100 == 0
       puts "Record count: #{index+1}"
-      sleep 10
+      sleep 15
     end
   end
 end
--- a/lib/tasks/import.rake
+++ b/lib/tasks/import.rake
@@ -1,0 +1,25 @@
+namespace :import do
+  task process_unfurled: :environment do
+    [
+      Imports::Doc,
+      Imports::Docx,
+      Imports::DocxListing,
+      Imports::Uncategorized
+    ].each do |model|
+      puts "Processing unfurled #{model} records"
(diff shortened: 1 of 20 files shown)


lib/tasks/import.rake
```diff
@@ -0,0 +1,26 @@
...
+      end
+
+      sleep 60
+    end
+  end
+
+  def rest(index)
+    if (index + 1) % 100 == 0
+      puts "Record count: #{index + 1}"
+      sleep 15
```
FerPerales, line 23: @littleforest do we still need this sleep?

lib/tasks/source_file.rake
```diff
@@ -0,0 +1,17 @@
...
+    count = 0
+    SourceFile.missing_import.find_each do |source_file|
+      import = source_file.create_import!
+
+      if import
+        count += 1
+
+        if count % 200 == 0
+          puts "Record count: #{index + 1}"
+          sleep 10
```
FerPerales, line 12: @littleforest same here. Not sure if this was only for debugging

FerPerales (approved):
Left some comments, but LGTM

lib/tasks/source_file.rake
littleforest: Just wanted to add a little sleep since we have a fair number of source files.

PR #804, opened 2024-05-28
Upgrade Administrate to 1.0.0beta

littleforest:
This upgrades Administrate to the 1.0 beta version. I removed most of the customized styling, since I don't think we are using it anymore. I added some styling fixes related to Administrate and Flowbite. There was also one styling issue with the Search bar which may be on the Administrate side, but fixing it on our side for now.

Screenshot 2024-05-28 at 12 05 37 PM

Screenshot 2024-05-28 at 12 07 33 PM


FerPerales (approved):
:shipit:

PR #814, opened 2024-05-31
Remove bulletin boolean from Standards Import model

littleforest:
With the new tree structure, the bulletin field on the StandardsImport model is no longer necessary. We pass the `listing: true` flag through from the Bulletins scraper job, which will correctly categorize the files.

[Asana ticket](https://app.asana.com/0/1203289004376659/1207208744528603)


FerPerales (approved):
This is great!

PR #816, opened 2024-06-04
Link document from RAPIDS API to import tree structure

littleforest:
When a document is available from the RAPIDS API, we create a StandardsImport record with an `Imports::Uncategorized` child containing the document, and process the record to obtain a PDF file. We then create a DataImport record linking the resulting pdf file and the `occupation_standard` that was created via the RAPIDS API, allowing the `data_import` file to be empty. Currently the RAPIDS API may have work process data, but no related instructions. If a converter is able to obtain the related instructions for the document at a later date, we want to be able to upload the related instructions through the existing job that processes data imports. To ensure that the work processes that were linked via the RAPIDS API are not deleted, we only destroy work processes if new work process data exists on the uploaded spreadsheet.

[Asana ticket](https://app.asana.com/0/1203289004376659/1207167427906520/f)


FerPerales (approved):
Thanks for working on this! LGTM

PR #842, opened 2024-06-18
Add Docs to admin section

FerPerales:
This PR only sets the skeleton for a Docs section in admin dashboard. Still not sure about the exact content we want there.

image

[Asana ticket](https://app.asana.com/0/1203289004376659/1207499629499036/f)


app/views/admin/docs/index.html.erb
```diff
@@ -0,0 +1,18 @@
...
+  <h1 class="main-content__page-title" id="page-title">
+    <%= content_for(:title) %>
+  </h1>
+</header>
+
+<section class="main-content__body main-content__body--flush">
+  <div class="main-content__page-description" style="margin-left:3rem;">
+    <dl>
+      <dt id="optionsForSearch">What are the options for search?</dt>
+      <dd>The available options depend on the tab</dd>
```
littleforest, line 15: The only interesting search that we have is on the Imports page, so you could add the information on how to filter on that page here (see examples in Asana ticket).

PR #854, opened 2024-06-21
Fix broken DataImport edit page

littleforest:
It seems like there is some
weirdness with the Administrate
has_many links sometimes being
nested and sometimes not. It will
be easier to investigate and clean
this up once the SourceFile has
been completely removed.

[Asana ticket](https://app.asana.com/0/1203289004376659/1207636165007789/f)

### DIFF
--- a/app/controllers/admin/data_imports_controller.rb
+++ b/app/controllers/admin/data_imports_controller.rb
@@ -64,10 +64,16 @@ module Admin
     private
 
     def set_parent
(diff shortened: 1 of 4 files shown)


FerPerales:
@littleforest looks good. Do we want to add some tests?

littleforest:
@FerPerales ready for re-review.

FerPerales (approved):

PR #867, opened 2024-06-26
Modify manual upload email for new imports feature

littleforest:
The email that goes out to guests who upload standards was not modified to handle the new import tree structure. This PR contains the changes for that. To reduce changes, the method names remain the same (using "source_file"), but the `Imports::Pdf` records are checked to see if they are completed and if a courtesy notification to the uploader is necessary.

This PR is best viewed hiding whitespace.

[Asana ticket](https://app.asana.com/0/1203289004376659/1207651928235988/f)

### DIFF
--- a/app/jobs/notify_uploaders_of_manual_conversion_completion_job.rb
+++ b/app/jobs/notify_uploaders_of_manual_conversion_completion_job.rb
@@ -2,18 +2,18 @@ class NotifyUploadersOfManualConversionCompletionJob < ApplicationJob
   queue_as :default
 
   def perform(email: nil)
-    StandardsImport.manual_submissions_in_need_of_courtesy_notification(email: email).each do |import|
+    StandardsImport.manual_submissions_in_need_of_courtesy_notification(email: email).each do |standards_import|
       GuestMailer.manual_upload_conversion_complete(
-        email: import.email,
-        source_files: import.source_files_in_need_of_notification
+        email: standards_import.email,
+        source_files: standards_import.source_files_in_need_of_notification
       ).deliver_now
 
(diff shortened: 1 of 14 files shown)


app/jobs/notify_uploaders_of_manual_conversion_completion_job.rb
```diff
@@ -2,18 +2,18 @@ class NotifyUploadersOfManualConversionCompletionJob < ApplicationJob
...
 
-      import.source_files_in_need_of_notification.each do |source_file|
+      standards_import.source_files_in_need_of_notification.each do |source_file|
         source_file.courtesy_notification_completed!
       end
 
-      if import.has_notified_uploader_of_all_conversions?
-        import.courtesy_notification_completed!
+      if standards_import.has_notified_uploader_of_all_conversions?
+        standards_import.courtesy_notification_completed!
```
littleforest, line 16: variable name change only since we now have an Import model.

app/models/standards_import.rb
```diff
@@ -10,12 +10,12 @@ class StandardsImport < ApplicationRecord
...
-      imports = StandardsImport.courtesy_notification_pending
+      standards_imports = StandardsImport.courtesy_notification_pending
       if email.present?
-        imports = imports.where(email: email)
+        standards_imports = standards_imports.where(email: email)
       end
-      imports.select do |import|
-        import.has_converted_source_file_in_need_of_notification?
+      standards_imports.select do |standards_import|
+        standards_import.has_converted_source_file_in_need_of_notification?
```
littleforest, line 18: variable name change only since we now have an Import model.

FerPerales (approved):

PR #870, opened 2024-06-26
Update link to public_document to use Imports::Pdf file

littleforest:
The "View Original Document" link is currently using the file attached to the StandardsImport, but those files are soon going to be removed. Use the file attached to the `Imports::Pdf` record instead for the new imports tree structure.

### DIFF
--- a/spec/views/occupation_standards/show_spec.rb
+++ b/spec/views/occupation_standards/show_spec.rb
@@ -1,0 +1,51 @@
+require "rails_helper"
+
+RSpec.describe "occupation_standards/show", type: :view do
+  context "without imports feature flag" do
+    it "when public document, it has link to original document" do
+      allow_any_instance_of(ActionView::TestCase::TestController).to receive(:controller_name).and_return("occupation_standards")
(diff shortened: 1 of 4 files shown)


jenlu33 (approved):

spec/views/occupation_standards/show_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+
+        render
+
+        expect(rendered).to have_text "View Original Document"
+      end
+    end
+  end
+
+  context "with imports feature flag" do
+    before do
```
FerPerales, line 27: @littleforest do we need this to be inside a `before` block? Seems we can move it inside the test and remove the block. Same with `after`

FerPerales (approved):

spec/views/occupation_standards/show_spec.rb
littleforest: Ripping out the feature flag soon, and this will make for less clean up than having to remove it from each separate test.

littleforest: (Actually only one test here, but there were multiple tests in another PR I did recently with same format)

PR #873, opened 2024-06-27
Make new imports structure permanent

littleforest:
This makes the new Imports tree structure permanent, and also removes the SourceFile code. This PR is best viewed hiding whitespace.

[Asana ticket](https://app.asana.com/0/1203289004376659/1207651928235983/f)
[Asana ticket](https://app.asana.com/0/1203289004376659/1207623200105842/f)


config/application.rb
```diff
@@ -56,8 +56,6 @@ class Application < Rails::Application
     # Rewrite FoxitSDK relative routes to redirect to /assets
     config.middleware.insert_before(Rack::Runtime, Rack::Rewrite) do
       r301 %r{lib/uix-addons/(.*)}, "/assets/lib/uix-addons/$1"
-      r301 %r{admin/source_files/([^/]*)/redact_file/lib/(.*)}, "/assets/lib/$2"
```
FerPerales, line 59: @littleforest heads up: I'll test in staging the PDF redaction feature since we don't have tests for the Foxit SDK plugin

lib/tasks/deployment/20230316205540_fix_changing_of_source_file_status_enum_value.rake
```diff
@@ -1,15 +0,0 @@
-namespace :after_party do
```
FerPerales, line 1: @littleforest is there any side effect if we delete this after party tasks? wondering if it would be best to keep them, but make them NOOP

FerPerales (approved):
@littleforest thanks for tackling this! I see that 95% was about removing the feature flag. Looks good. Left a couple of comments

config/application.rb
littleforest: Great, thanks!

lib/tasks/deployment/20230316205540_fix_changing_of_source_file_status_enum_value.rake
littleforest: No, there's no issue. They've already run on prod, and would just get run in dev by a new developer, so doesn't matter if they are missing or noop. We have the task id in the `task_versions` table, so we could always use it to find the file in git if needed (seems unlikely though).

PR #883, opened 2024-07-11
Archive extracted bulletin files

jenlu33:
[Asana ticket](https://app.asana.com/0/1203289004376659/1207439838600406/f)

I created a service that is called after a PDF is created that checks the file if it matches certain criteria and then archives it if necessary. Using PDF Reader and then feeding that text into ChatGPT is giving consistent results vs just uploading directly to ChatGPT did not give consistent results


spec/services/pdf_archive_evaluator_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+      ).and_return chat_gpt_generator_mock("true")
+
+      expect { service.call }.not_to change(pdf, :status)
+    end
+  end
+end
+
+def chat_gpt_generator_mock(value)
+  OpenStruct.new(call: value)
+end
```
littleforest, line 56: nice!

littleforest (commented):
@jenlu33 looks really good to me! Mostly just styling suggestions.

littleforest (commented):
Also, what do you think about making this a background job instead of a service? I think it makes sense to have it be a background job since nothing is really dependent on it and it can run separately. It should also fix the failing specs that are now trying to call the service.

FerPerales:
@jenlu33 since we now have a plan for the OpenAI extraction, I'll close this.

PR #900, opened 2024-09-03
The system prompts someone to provide their name, email, and organization after viewing two standards

FerPerales:
[Asana ticket](https://app.asana.com/0/1203289004376659/1205288992536752/f)

Since we rely on cookies, we first ask the user if it's OK to use them:

image

If user accepts, we show the modal based on the behavior described:

Modal:

image

If they don't accept, they will never see the modal.

Admins can see all the responses submitted by users: 

image


jenlu33 (approved):
Looks good to me once all tests pass

app/policies/survey_policy.rb
```diff
@@ -0,0 +1,2 @@
+class SurveyPolicy < AdminPolicy
```
clarissalimab, line 1: Are you planning to use this in the future? Or is this automatically generated?

clarissalimab (approved):
Looks good to me!

PR #920, opened 2024-10-07
Store pdf text on elasticsearch

FerPerales:
[Asana ticket](https://app.asana.com/0/1203289004376659/1208347992671016/f)

Let's store the PDF text into PostgreSQL. We may not need to store it in ES to do the text substitution


jenlu33 (approved):
Looks good to me

PR #948, opened 2024-10-28
rough in query to find standards imports to be included in monthly email

rdnewmanbot:
Asana: [The system can send a monthly notice of newly converted standards](https://app.asana.com/0/1203289004376659/1205982289180734)


spec/models/standards_import_spec.rb
```diff
@@ -47,29 +47,6 @@
     expect(import.reload.name).to eq "Mickey Mouse"
   end
 
-  describe "imports" do
```
rdnewmanbot, line 50: only moving down to be below class method specs

app/models/import.rb
```diff
@@ -21,11 +21,11 @@ class Import < ApplicationRecord
     :needs_backend_support,
     :unfurled
   ]
-  enum courtesy_notification: [
+  enum :courtesy_notification, [
```
rdnewmanbot, line 24: This and similar `enum` changes are to address a syntax deprecation warning that we see during rspec runs.

spec/system/admin/data_imports/new_spec.rb
```diff
@@ -11,7 +11,7 @@
     within("h1") do
       expect(page).to have_content("Import data for pixel1x1.pdf")
     end
-    expect(page).to have_alert, text: "Read Carefully"
```
rdnewmanbot, line 14: Addresses syntax warning seen during rspec runs

PR #952, opened 2024-10-29
Detect words in TextRepresentation#content from WordReplacement list

FerPerales:
[Asana ticket](https://app.asana.com/0/1203289004376659/1208347992671019/f)

Each OccupationStandard has a related document (Import) with all the details about itself. Some of those documents contain words that we want to replace since they are either inappropriate or outdated.

First part of this feature is detect if an OccupationStandard has a document with those words. Each OccupationStandard has a TextRepresentation model, where we store the contents of its corresponding Import.

We have a new model called WordReplacement, where we store the list of words we want to replace. For now, we don't have an official list, but admins can add terms manually.

This PR introduces a query that checks if, given a text_representation_id, its contents have one of the words we want to replace.

UI for this is TBD

Additional notes:

This complete feature is a PoC to proof that this can be achieved. If we decide to implement this as a long-term supported feature, we may want to create a index to optimize the search process and use the index instead of asking postgreSQL to parse the content each time.  

```
CREATE INDEX text_representation_content_idx ON text_representation USING GIN (to_tsvector('english', content));
```

### DIFF
--- a/app/queries/detect_disallowed_words_in_text_representation_query.rb
+++ b/app/queries/detect_disallowed_words_in_text_representation_query.rb
@@ -1,0 +1,27 @@
+class DetectDisallowedWordsInTextRepresentationQuery
+  def self.run(text_representation_id)
+    result = TextRepresentation.connection.exec_query(
+      TextRepresentation.sanitize_sql([
+        "
+          SELECT array_agg(rm[1]) AS WORDS
+            FROM (
+              SELECT ts_headline(
+                content,
+                to_tsquery(array_to_string(array_agg(word_replacements.word), ' | '))
+              )
+              FROM text_representations
+              CROSS JOIN word_replacements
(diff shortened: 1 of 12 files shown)


app/queries/detect_disallowed_words_in_text_representation_query.rb
```diff
@@ -0,0 +1,27 @@
+class DetectDisallowedWordsInTextRepresentationQuery
+  def self.run(text_representation_id)
+    result = TextRepresentation.connection.exec_query(
+      TextRepresentation.sanitize_sql([
+        "
+          SELECT array_agg(rm[1]) AS WORDS
```
rdnewmanbot, line 6: This was good -- I learned some new things. Thank you!

rdnewmanbot (approved):
Looks good!

PR #966, opened 2024-11-14
Clarify invitation email in user creation flash message and add tests for invite action

f-moya:
This change aims to reduce potential confusion regarding the invitation email sent after user creation through the admin panel. The flash message now explicitly mentions that an invitation email has been sent.

Additionally, introduce tests for the `Admin::UsersController#invite` action, which previously lacked test coverage. This is the first in a series of changes related to this area, as outlined in the attached Asana ticket.

[Asana ticket](https://app.asana.com/0/1203289004376659/1206201273701899/f)

### DIFF
--- a/app/controllers/admin/users_controller.rb
+++ b/app/controllers/admin/users_controller.rb
@@ -17,7 +17,7 @@ module Admin
         resource.invite!
         redirect_to(
           after_resource_created_path(resource),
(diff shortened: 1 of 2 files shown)


FerPerales (approved):
Looks good! Thanks for putting this together!

PR #997, opened 2024-12-11
Update enums to use positional arguments

clarissalimab:
Defining enums with keyword arguments is deprecated and will be removed in Rails 8.0.

[Asana ticket](https://app.asana.com/0/1203289004376659/1208946454714575)

### DIFF
--- a/app/models/import.rb
+++ b/app/models/import.rb
@@ -21,11 +21,11 @@ class Import < ApplicationRecord
     :needs_backend_support,
     :unfurled
   ]
-  enum courtesy_notification: [
+  enum :courtesy_notification, [
     :not_required,
     :pending,
     :completed
-  ], _prefix: true
+  ], :prefix => true
 
   scope :needs_unfurling, -> { unfurled.where("created_at < ?", 1.day.ago) }
 
--- a/app/models/import.rb
+++ b/app/models/import.rb
@@ -25,7 +25,7 @@ class Import < ApplicationRecord
     :not_required,
     :pending,
     :completed
-  ], :prefix => true
+  ], prefix: true
 
   scope :needs_unfurling, -> { unfurled.where("created_at < ?", 1.day.ago) }
 
--- a/app/models/occupation_standard.rb
+++ b/app/models/occupation_standard.rb
@@ -23,8 +23,8 @@ class OccupationStandard < ApplicationRecord
   delegate :name, to: :industry, prefix: true, allow_nil: true
   delegate :state, to: :registration_agency, allow_nil: true
 
-  enum :ojt_type, [:time, :competency, :hybrid], :suffix => :based
-  enum :national_standard_type, [:program_standard, :guideline_standard, :occupational_framework], :prefix => :national
+  enum :ojt_type, [:time, :competency, :hybrid], suffix: :based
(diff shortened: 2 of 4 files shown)


app/models/import.rb
```diff
@@ -21,11 +21,11 @@ class Import < ApplicationRecord
     :needs_backend_support,
     :unfurled
   ]
-  enum courtesy_notification: [
+  enum :courtesy_notification, [
     :not_required,
     :pending,
     :completed
-  ], _prefix: true
+  ], :prefix => true
```
slickepinne, line 28: I think you can write

```suggestion
  ], prefix: true
```
instead or is there a reason why we use the old syntax?

app/models/occupation_standard.rb
```diff
@@ -23,9 +23,9 @@ class OccupationStandard < ApplicationRecord
   delegate :name, to: :industry, prefix: true, allow_nil: true
   delegate :state, to: :registration_agency, allow_nil: true
 
-  enum ojt_type: [:time, :competency, :hybrid], _suffix: :based
-  enum national_standard_type: [:program_standard, :guideline_standard, :occupational_framework], _prefix: :national
-  enum status: [:importing, :in_review, :published]
+  enum :ojt_type, [:time, :competency, :hybrid], :suffix => :based
```
slickepinne, line 26: Same here. We could use `suffix: :based` and `prefix: :national`. You can see more examples in the [ActiveRecord::Enum documentation for Rails 8.01](https://api.rubyonrails.org/classes/ActiveRecord/Enum.html). But again, maybe there's a reason we are not using it. ☺️

app/models/standards_import.rb
```diff
@@ -1,7 +1,7 @@
 class StandardsImport < ApplicationRecord
   has_many :imports, as: :parent, dependent: :destroy
 
-  enum courtesy_notification: [:not_required, :pending, :completed], _prefix: true
+  enum :courtesy_notification, [:not_required, :pending, :completed], :prefix => true
```
slickepinne, line 4:
```suggestion
  enum :courtesy_notification, [:not_required, :pending, :completed], prefix: true
```

slickepinne (commented):
Nice one. I'm happy that we take care of deprecation warnings. I just left a few comments regarding the syntax. 🙂

app/models/import.rb
clarissalimab: You're right, that was my old Ruby brain 😅 Thanks for catching that.

app/models/occupation_standard.rb
```diff
@@ -23,9 +23,9 @@ class OccupationStandard < ApplicationRecord
...
+  enum :national_standard_type, [:program_standard, :guideline_standard, :occupational_framework], :prefix => :national
+  enum :status, [:importing, :in_review, :published]
```
clarissalimab, line 28:
```suggestion
  enum :ojt_type, [:time, :competency, :hybrid], suffix: :based
  enum :national_standard_type, [:program_standard, :guideline_standard, :occupational_framework], prefix: :national
  enum :status, [:importing, :in_review, :published]
```

FerPerales (approved):

PR #1002, opened 2024-12-17
Add dynamic work processes field to occupation standard

slickepinne:
Occupation Standards have many work processes. When we edit an occupation standard, we want to dynamically add and remove work processes without changing dashboards. During a first iteration we thought of using the rondo_form gem but since we are using administrate, we would need to go against rails conventions to make it work. The [administrate-field-nested_has_many gem](https://github.com/nickcharlton/administrate-field-nested_has_many) provides exactly the workflow we are looking for so I would suggest to move forward with this one instead of rondo_form.

This PR only adds the functionality to add and remove a work process. We will update the UI in a follow up PR.

### Visualisation

Dynamic_Work_Processes_on_Occupation_Standards

[Asana ticket](https://app.asana.com/0/1203289004376659/1208975678543527/f)


clarissalimab:
@slickepinne, this is great! My only concern with the way it currently is is that if a record has many work processes, it might be a very long view (especially because we're planning to also nest the competencies on each one of these) - I wonder if there is a way to make it expandable and potentially paginate it. I'll look into that :)

app/dashboards/work_process_dashboard.rb
```diff
@@ -40,18 +40,16 @@ class WorkProcessDashboard < Administrate::BaseDashboard
...
-    competencies
-    default_hours
+    title
     description
+    default_hours
     maximum_hours
     minimum_hours
-    sort_order
-    title
   ].freeze
```
clarissalimab, line 48: This change is affecting the existing edit page for the work processes. It's interesting to note that the current behavior takes forever to load because of the competencies.

FerPerales (approved):
It's perfectly fine to use an existing gem. Looks good! Thanks for working on this!

app/dashboards/work_process_dashboard.rb
FerPerales: @clarissalimab can you add a new task to not forget about this? thanks!

clarissalimab: Before the change:
image

After the change:
image

slickepinne: This is resolved as I merged #1006 into this PR.

FerPerales (approved):

clarissalimab (approved):

PR #1010, opened 2024-12-19
Update UI for nested has_many fields

slickepinne:
An occupation standard can have multiple work processes. While we limit the display of these work processes on the index page, we showed all of them, including their competencies, when we edited an occupation standard. It was really difficult to navigate through these fields. In order to make it easier, we used details and summaries to show/hide the work processes and competencies.

### Visualisation
Screenshot 2024-12-19 at 11 33 12

[Asana ticket - Subtask](https://app.asana.com/0/1203289004376659/1208975678543527/f)


clarissalimab (approved):
Looks good to me! I really like the approach you took :)

FerPerales (approved):
Love it!

PR #1014, opened 2024-12-24
Fix admin styling

slickepinne:
Table content did not get displayed correctly on
smaller devices. The table cells also did not
wrap nicely and the Filter By button wrapped the
SVG to the next line. This commit fixes all
these little issues.

The input search field does not turn blue on hover
any longer, so I removed the customised styling.

### Visualisation

**Before**
Import_Before

**After**
Import_after


FerPerales (approved):
Thanks for fixing this!

PR #1022, opened 2024-12-30
Fix Claim button

slickepinne:
With the last upgrade of administrate, turbo got
introduced. The form for the claim button on imports is nested in an anchor tag. With turbo enabled on links by default, the link itself takes precedence over the form. This led to the fact that we always got redirected to the show page of an import but we were not able to add an assignee. This commit fixes it.

[Asana ticket](https://app.asana.com/0/1203289004376659/1209041701094994/f)


rdnewmanbot (approved):
Looks great -- nice sleuthing!

PR #1033, opened 2025-01-14
Populate work processes from OpenAI

FerPerales:
[Asana ticket](https://app.asana.com/0/1203289004376659/1209041701094991/f)

### DIFF
--- a/app/services/generate_training_data.rb
+++ b/app/services/generate_training_data.rb
@@ -9,7 +9,7 @@ class GenerateTrainingData
   def call
     response = @client.chat(
       parameters: {
-        model: "gpt-4",
+        model: "gpt-4-mini",
         messages: [{role: "user", content: prompt}],
         temperature: 0.2
       }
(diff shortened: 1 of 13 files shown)


app/services/generate_training_data.rb
```diff
@@ -9,7 +9,7 @@ def initialize(text)
   def call
     response = @client.chat(
       parameters: {
-        model: "gpt-4",
+        model: "gpt-4-mini",
```
clarissalimab, line 12: Should this match the other files?

```suggestion
        model: "gpt-4o-mini",
```

clarissalimab (approved):
Looks good to me! Nice design on the JsonImportable concern.

app/services/generate_training_data.rb
FerPerales: @clarissalimab good catch! it should be `gpt-4o-mini`

PR #1043, opened 2025-01-23
Store OpenAI response

FerPerales:
[Asana ticket](https://app.asana.com/0/1203289004376659/1209203103872886/f)

This PR introduces a model called `OpenAIImport`, which stores the response we got from OpenAI and the import_id used for the convertion

This model will be used to evaluate the difference between the OpenAI response and the final version saved by editors.


clarissalimab (approved):
Looks good!

PR #1069, opened 2025-02-18
Fetch OpenAI data as a background job and notify converter when it's ready to convert

FerPerales:
[Asana ticket](https://app.asana.com/0/1203289004376659/1209378338454321)

Run data extraction as a background job and notify converters when document is ready for review


app/controllers/admin/occupation_standards_controller.rb
```diff
@@ -1,16 +1,22 @@
...
-      }
+        occupation_standard.open_ai_response = @open_ai_import.response
+        occupation_standard.import_id = params[:import_id]
+        authorize_resource(occupation_standard)
+        render locals: {
+          page: Administrate::Page::Form.new(dashboard, occupation_standard)
+        }
+      else
+        skip_authorization
+        redirect_to admin_import_path(params[:import_id]), notice: "You must prepare the document by clicking Convert with AI"
```
clarissalimab, line 18: Just out of curiosity, since it seems this is mandatory now, is there a reason for not directly triggering the Import? Is it because it can take a while to do that?

clarissalimab (approved):
Looks good to me!

app/controllers/admin/occupation_standards_controller.rb
FerPerales: @clarissalimab correct! We will also want to experiment with OpenAI batch processing, which means a petition may not get resolved for hours, but it will be cheaper. We don't really need a response in real time

PR #1098, opened 2026-05-19
Create sample of 500 Occupation Standards records

ckuttruff:
See the following Asana ticket for background:
https://app.asana.com/1/33342061919290/project/1202142120438783/task/1214504264358398?focus=true

## Changes
* adds boolean flag on occupation_standards table to differentiate records
* creates sample selection class for specifying criteria for samples
* adds rake task to run the above class and set this sample flag
* augments the occupation standards detail page to allow the user to filter by this flag


app/controllers/admin/application_controller.rb
```diff
@@ -10,6 +10,11 @@ class ApplicationController < Administrate::ApplicationController
 
     def authenticate_admin
       authenticate_user!
+
+      return if current_user.admin? || current_user.converter?
```
Pcushing, line 15: I assumed that `after_action :verify_authorized # Pundit` did this part? It must not. I havent used pundit before.

Either way, i'd rather pull this check out from the core `authenticate_admin` and do another `before_action` like:

`before_action :signout_user, if: proc { current_user.admin? || current_user.converter? } `

And let that separately handle the signout + redirect.

app/dashboards/occupation_standard_dashboard.rb
```diff
@@ -21,6 +21,7 @@ class OccupationStandardDashboard < Administrate::BaseDashboard
     related_job_titles: Field::String.with_options(searchable: false),
     rsi_hours_max: Field::Number,
     rsi_hours_min: Field::Number,
+    sample: Field::Boolean,
```
Pcushing: Nit pick, but can we rename to `sample_set` or something that sounds a bit more like a label than a verb?

app/services/occupation_standard_sample_selector.rb
```diff
@@ -0,0 +1,131 @@
+require "zlib"
+
+class OccupationStandardSampleSelector
```
Pcushing, line 3: This is honestly cleaner than it had to be for code that'll really be used 1x!

.gitignore
```diff
@@ -36,6 +36,9 @@
 !/app/assets/builds/.keep
 
 .env
+.envrc
+latest.dump
```
Pcushing, line 40: It's very likely that we'll want to pull the `latest.dump` for local work periodically given the data set here. It might be helpful to have a script to do so in the project. There's still an old version of this in the workhands project for reference when we were on Heroku - me_want_workhands_db.sh

Pcushing (commented):
Just a little tidying but it looks like there's some test infra issues, too.

app/controllers/admin/application_controller.rb
ckuttruff: Updated, thanks

app/dashboards/occupation_standard_dashboard.rb
ckuttruff: yeah, that's a clearer name; updated that and re-deploying to staging, thanks

.gitignore
ckuttruff: yeah, I'll add a ticket for myself for this. There is a secondary issue of the corresponding assets (since that prod db data will be referring to pdfs / docs that only exist in the prod bucket and we don't want to reference that locally), but let's chat about those nuances tomorrow

ckuttruff:
> Just a little tidying but it looks like there's some test infra issues, too.

@Pcushing yeah had noticed that failure; was tweaking database.yml to accommodate for Unix domain sockets (which Linux uses by default for postgres connections), but had to tweak how CI was connecting since it's still relying on TCP for the Github Actions workflow.

I've adjusted that and it looks like tests are running properly now (it is taking quite a long time so that's another thing to revisit, because 15+ minutes to run test suite seems not ideal)

Pcushing (approved):

PR #1103, opened 2026-07-13
Security updates for various gems

ckuttruff:
[Asana ticket](https://app.asana.com/1/33342061919290/project/1202142120438783/task/1216495431607398?focus=true)


app/resources/api/v1/occupation_resource.rb
```diff
@@ -1,6 +1,7 @@
 class API::V1::OccupationResource < JSONAPI::Resource
   immutable
   model_name "Occupation"
+  self._routed = true
```
ckuttruff, line 4: fix necessary for the rails upgrade and how JSON API routes

config/initializers/administrate.rb
```diff
@@ -1 +1,19 @@
 Administrate::Engine.add_stylesheet("administrate_customizations")
+
+require "administrate/search"
```
ckuttruff, line 3: patch to fix an annoying deprecation notice since the rails 8.1 upgrade. Worth submitting an upstream PR later, but patch is for expediency in the immediate term

ckuttruff:
@Pcushing this is out on staging and things are looking good on there and also localhost; lemme know if you have any feedback but I'm going to go ahead and merge

